PR #10, opened 2020-11-24
#2 - Change project structure

JustasBelevicius:


src/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "tsc && node dist/index.js",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "debug": "tsc && nodemon dist/index.js"
+    "debug": "tsc && nodemon dist/index.js -e js,ts"
```
ignasposka, line 9: These parameters are already put into `nodemon.json`. Btw, I don't think it should watch for js files as we write only in ts.

ignasposka (approved):
Works like a charm!

PR #19, opened 2020-12-09
Set up debugging.

ignasposka:


AzisK (approved):
Looks good to me (though I am not an expert in this matter)

PR #23, opened 2020-12-15
Updating ER diagram

ignasposka:


psaulenas:
Maybe would be possible to also add a file that would be editable?

ignasposka:
> Maybe would be possible to also add a file that would be editable?

I will try to get it from LucidChart

ignasposka:
@psaulenas I cannot export editable file, because it's a premium feature...

AzisK:
I suppose we can make the country an ENUM since there is a known number of countries in the world. Either ISO-2 or ISO-3 code or even the full country name. What do you think?

AzisK (changes requested):
- app_user mod_time is missing NOT NULL
- organization id was changed from SMALLSERIAL to SERIAL based on https://github.com/pets-oss/pets-back/pull/17#discussion_r537478494
- registration_date is missing in animal according to https://github.com/pets-oss/pets-back/pull/17#issuecomment-740102577

ignasposka:
> I suppose we can make the country an ENUM since there is a known number of countries in the world. Either ISO-2 or ISO-3 code or even the full country name. What do you think?

Yes, we can do it. I will update diagram.

AzisK (approved):
Good job! We also have NOT NULL next to registration_date but I am not sure how that should look like in ERD

Issue #21, opened 2020-12-15
Improve database


AzisK:
Make foreign keys NOT NULL

and other

AzisK:
Also
- make primary keys NOT NULL
- convert breed to an ENUM

PR #24, opened 2020-12-17
Improve database #21

AzisK:
From https://github.com/pets-oss/pets-back/issues/21

- make foreign keys NOT NULL
- make primary keys NOT NULL
- convert breed to an ENUM

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -1,5 +1,5 @@
 CREATE TABLE organization (
-    id SERIAL PRIMARY KEY,
+    id SERIAL PRIMARY KEY NOT NULL,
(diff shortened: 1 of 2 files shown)


database/schema.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE organization (
-    id SERIAL PRIMARY KEY,
+    id SERIAL PRIMARY KEY NOT NULL,
```
psaulenas, line 2: PRIMARY KEY has unique and NOT NULL constraints, so it is not needed to add it

ignasposka (approved):

psaulenas (approved):
LGTM

PR #28, opened 2020-12-22
#8 Setup proper heroku deployment

JustasBelevicius:
- Runs the app on `https://petbook-back.herokuapp.com/graphiql`
- Postgres db had to be initialised by hand, I might create an action later, but it would have to drop all tables before updating - loosing all data we had. For now, I believe it is better to have it be difficult to update, to avoid accidental data drops. Might want to eventually have some kind of `database/updates/1.sql` files to update the tables without loosing data.


JustasBelevicius:
Note! We don't want to run the Deploy Heroku job on pull_request open :D

src/services/postgres.ts
```diff
@@ -29,13 +30,23 @@ const pgClient = (pool: Pool): PostgresClient => ({
...
+        },
+      }
+    : {
+        user: process.env.POSTGRES_USER,
+        host: process.env.POSTGRES_HOST,
+        database: process.env.POSTGRES_DB,
+        password: process.env.POSTGRES_PASSWORD,
+        port: parseInt(process.env.POSTGRES_PORT || '5432', 10),
+      };
+  console.log(config);
```
psaulenas, line 48: Could you please remove `console.log`?

package.json
```diff
@@ -4,11 +4,13 @@
...
   "scripts": {
-    "start": "tsc && nodemon dist/index.js",
+    "start": "nodemon dist/index.js",
     "test": "echo \"Error: no test specified\" && exit 1",
     "debug": "tsc && nodemon --inspect-brk=0.0.0.0 dist/index.js",
     "check-lint": "eslint 'src/**/*.ts'",
-    "check-type": "tsc --noEmit --skipLibCheck"
+    "check-type": "tsc --noEmit --skipLibCheck",
+    "build": "tsc",
+    "start-prod": "npm run build && npm run start"
```
psaulenas, line 13: Why did you also create `start-prod`?

psaulenas (approved):
GJ! Left a few comments

src/services/postgres.ts
```diff
@@ -1,3 +1,4 @@
+import console from 'console';
```
psaulenas, line 1: Seems not needed import

JustasBelevicius: Will add a rule to husky if possible

JustasBelevicius: Looks like the current `eslint` setup checks for unused imports. This one is used in the PostgresClient disconnect callback, if an error is caught.

AzisK (approved):
Looks smooth, good job

src/services/postgres.ts
psaulenas: My guess would be that your IDE did auto-import. Removing this import shouldn't affect anything, you don't need to import `console`

ignasposka (approved):

src/services/postgres.ts
JustasBelevicius: That's exactly what happened :D

PR #31, opened 2020-12-30
26 add real types

ignasposka:
Added types according to http://archyvas.vic.lt/?mid=738. Using only IDs in enum types, because looks like IDs are important in this classification system. Added translation tables linked to enums to get type name in LT language (it will be shown in frontend).
I have also dropped tables `animal_event_check_in` and `animal_event_check_out` as they look redundant when `animal_event_general` exists.


database/3-add_types.sql
```diff
@@ -0,0 +1,1110 @@
...
+UPDATE animal SET species = '2' WHERE species = 'snake';
+UPDATE animal SET species = '3' WHERE species = 'hamster';
+ALTER TYPE species RENAME TO species_old;
+
+CREATE TYPE species AS ENUM ('1', '2', '3');
+ALTER TABLE animal ALTER COLUMN species TYPE species USING species::text::species;
+DROP TYPE species_old;
+
+CREATE TABLE species_translation (
+    species species PRIMARY KEY,
```
psaulenas, line 18: `species` can't be PK, we wouldn't support then multiple languages

database/3-add_types.sql
```diff
@@ -0,0 +1,1110 @@
...
+
+-- Changing breed to table to hold id and code values
+ALTER TABLE animal DROP COLUMN breed;
+DROP type breed;
+
+CREATE TABLE breed (
+    id INTEGER PRIMARY KEY,
+    code VARCHAR(7),
+    species species,
+    UNIQUE (id, code)
```
psaulenas, line 38: I think it would be enough to have additional `UNIQUE` constraint only on `code` column because `id` is already PK and has this constraint. Also, this would add additional safety to not have duplicates for `code` values.

psaulenas (commented):
Left a few questions and comments

database/3-add_types.sql
ignasposka: Good point, thanks!

database/3-add_types.sql
```diff
@@ -0,0 +1,1110 @@
...
+UPDATE animal SET species = '1' WHERE species = 'dog';
+UPDATE animal SET species = '2' WHERE species = 'snake';
+UPDATE animal SET species = '3' WHERE species = 'hamster';
+ALTER TYPE species RENAME TO species_old;
+
+CREATE TYPE species AS ENUM ('1', '2', '3');
+ALTER TABLE animal ALTER COLUMN species TYPE species USING species::text::species;
+DROP TYPE species_old;
+
+CREATE TABLE species_translation (
```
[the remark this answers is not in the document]
ignasposka: 1. I have a long list of breeds in Lithuanian language. Therefore I have two options:
a. Translate everything manually into English and put in the database
b. Put Lithuanian translations into database with idea that English is not present at the moment

For our current users we are going to show Lithuanian values, so there is no point at the moment to translate all values to English. Therefore I have chosen second option.

2. I could agree with that. I can move translations to JSON file.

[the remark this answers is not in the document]
JustasBelevicius: 2. Having translations in some json file in the project would require the project to be redeployed. Currently the app redeploys on any file change, so the SQL change would cause unneccessary app redeploy anyways. In the near future we would have to separate actions for SQL and app deploy, currently SQL is deployed by hand for security reasons.

JustasBelevicius (approved):

database/3-add_types.sql
```diff
@@ -0,0 +1,565 @@
+-- Adding species types
+ALTER TYPE species ADD VALUE '1';
```
AzisK, line 2: Since the app was not running in production and no production data was added, I believe we do not necessarily need to add changes as SQL-change-code. To simplify our code and SQL commits, I recommend to just change the underlying schema and data SQLs

AzisK (commented):
Good job overall. I suggest to keep it simple and not alter or drop tables, columns but delete them from initial schema and data SQLs

database/3-add_types.sql
ignasposka: Hm, I tried to keep it incremental and do not touch earlier written files. Do you think it's worth rewriting now?

AzisK: I believe that's a good approach but now I suggest to keep it simple stupid and reduce complexity as much as possible as in https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#general-rules We will have people joining this code, so I believe it is much easier to understand the schema when it is all in one place and not scattered around. I believe it's worth it. I am sorry for not bringing this earlier

psaulenas (approved):
LGTM

database/3-add_types.sql
```diff
@@ -34,8 +35,7 @@ DROP type breed;
 CREATE TABLE breed (
     id INTEGER PRIMARY KEY,
     code VARCHAR(7),
-    species species,
-    UNIQUE code
```
psaulenas: Did you remove this constraint on purpose?

ignasposka: Yes. It turns out code is not unique, cat and dog can share same code.

database/3-add_types.sql
ignasposka: Initial SQL will almost never display current schema, because sooner or later changes will be introduced and by time passing initial SQL will be more and more out of sync.

I'm not against your idea, but this time I would choose to keep as it is because of time shortage (this PR blocks other tasks), furthermore both solutions have advantages and disadvantages.

database/3-add_types.sql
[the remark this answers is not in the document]
ignasposka: Choosing SQL approach for more seamless relation between data and translations.

PR #43, opened 2021-01-07
Changed database structure

JBK91:
- Removed `organization` column from `event` tables
- Created `app_user_role` table - one user may have role in more than 1 organisation
- Removed street address slitting from `organization` table - IMO we don't need 3 columns for street address (street, house and flat)
- created `animal_registration` table to store registration data and added `status` column to track if that registration is active
- created `animal_microchip` table to store microchip data and added `status` column to track if that chip is implanted
- created `animal_details` table to split animal general info from `animal` table

### DIFF
--- a/database/4-change_structure.sql
+++ b/database/4-change_structure.sql
@@ -1,0 +1,101 @@
+ALTER TABLE animal_event_general
+DROP COLUMN organization;
+
+ALTER TABLE animal_event_medical_record
+DROP COLUMN organization;
+
+ALTER TABLE app_user DROP COLUMN organization,
+DROP COLUMN role_type;
+
+CREATE TABLE app_user_roles (
+    user_id VARCHAR(256) REFERENCES app_user(id) NOT NULL,
+    organization_id INTEGER REFERENCES organization(id) NOT NULL,
+    role_type role_type DEFAULT 'Guest',
+    PRIMARY KEY (user_id, organization_id)
+);
+
+INSERT INTO app_user_roles (user_id, organization_id, role_type) VALUES
+    ('aiubfaw4io09', 1, 'Member'),
+    ('afhu9w4f78', 2, 'Owner'),
+    ('278y2378ryb', 2, 'Member'),
+    ('0932hfdsa', 2, DEFAULT),
+    ('dhjbwau74a6', 1, 'Owner');
+
+ALTER TABLE organization
+DROP COLUMN street,
+DROP COLUMN house,
+DROP COLUMN flat,
+ADD COLUMN street_address VARCHAR(256);
+
+UPDATE organization SET street_address = 'Mindaugo 12-1' WHERE id = 2;
+UPDATE organization SET street_address = 'Ulrich Strasse 99' WHERE id = 3;
+UPDATE organization SET street_address = 'Kalnų 38-5' WHERE id = 4;
+UPDATE organization SET street_address = 'Vytauto 1a' WHERE id = 1;
+
+CREATE TYPE registration_status AS ENUM ('Active', 'Inactive');
+
+CREATE TABLE animal_registration (
+    animal_id INTEGER REFERENCES animal(id),
+    registration_no VARCHAR(256) NOT NULL,
+    registration_date DATE NOT NULL,
+    status registration_status DEFAULT 'Active',
+    PRIMARY KEY (animal_id, registration_no)
+);
+
+INSERT INTO animal_registration (animal_id, registration_no, registration_date, status) VALUES
+    (1, '123Svx', '2021-01-07', DEFAULT),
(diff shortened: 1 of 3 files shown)


JustasBelevicius (approved):
Well done 👍 ! Since the product is not in production stages yet, it would be better to add all the schema changes to the 1-schema.sql editing the create queries directly. I know it wasn't done like that before, but I believe that would look cleaner until we are forced to do it all with alterations when the app is used.

tl;dr; 
- CREATE and ALTER (just edit table CREATION queries instead) can be moved to the 1-schema.sql
- INSERT queries go to 2-data.sql, it won't be run on production in the later stages of the app.

AzisK (approved):
Good job!

AzisK:
> Well done 👍 ! Since the product is not in production stages yet, it would be better to add all the schema changes to the 1-schema.sql editing the create queries directly. I know it wasn't done like that before, but I believe that would look cleaner until we are forced to do it all with alterations when the app is used.

We can do this with a separate pull request since this approach is already used before

database/4-change_structure.sql
```diff
@@ -0,0 +1,101 @@
...
+);
+
+INSERT INTO animal_details (animal_id, breed_id, name, gender, color, birth_date, weight, allergy, food) VALUES
+    (1, 205, 'Haskelis', 1, 61, '2020-08-01', 20, NULL, NULL),
+    (2, 268, 'Jupyteris', 1, 2, '2020-01-01', 1, NULL, NULL),
+    (5, 389, 'Murkė', 2, 4, '2019-09-16', 7, NULL, NULL),
+    (4, 350, 'Javainis', 1, 68, '2020-01-01', 0.4, NULL, NULL),
+    (3, 422, 'Pitoncas', 1, 32, '2020-01-01', 0.3, NULL, NULL);
+
+ALTER TABLE animal
```
psaulenas, line 88: Please update also https://github.com/pets-oss/pets-back/blob/main/src/schemas/animal.ts#L28

psaulenas (changes requested):
You need also update GraphQL schema

database/4-change_structure.sql
ignasposka: Why animal GQL schema should be updated? I think it should remain the same for the user of GQL endpoint 🤔

ignasposka (changes requested):
I'm not sure what we are doing with animal type in GraphQL side. At this point I see removing lot of fields from animal type, but they still somehow should be available through GQL endpoint, what's the plan for that?

database/4-change_structure.sql
JBK91: `created animal_registration table to store registration data and added status column to track if that registration is active
created animal_microchip table to store microchip data and added status column to track if that chip is implanted`

in the rare case microchips can be changed (damaged or something), registrations too. With these tables we will have historical data for animal. If that data will be only in animal where can be only one value, we lose historical data.

`animal_details` is to split general info/details from `animal` table

JBK91:
> I'm not sure what we are doing with animal type in GraphQL side. At this point I see removing lot of fields from animal type, but they still somehow should be available through GQL endpoint, what's the plan for that?

I am not sure what needs for that, if frontend needs to have all data calling `animal`, than I will update animal query to include other related data, in other cases I will add endpoints.

IMO: `animal_details` could be in `animal` endpoint:
{
   animal {
         ....,
         details {
              ....
         }
   }
}

other (microchips and registrations) other endpoints

ignasposka:
> > I'm not sure what we are doing with animal type in GraphQL side. At this point I see removing lot of fields from animal type, but they still somehow should be available through GQL endpoint, what's the plan for that?
> 
> I am not sure what needs for that, if frontend needs to have all data calling `animal`, than I will update animal query to include other related data, in other cases I will add endpoints.
> 
> IMO: `animal_details` could be in `animal` endpoint:
> {
> animal {
> ....,
> details {
> ....
> }
> }
> }
> 
> other (microchips and registrations) other endpoints

I think we could have GQL query `Animal` which would let to get animal with all microchips, details and registrations. This is the one of the main GQL feature, to let GQL client choose exactly what it wants. It would be silly to not use this advantage of GQL. In my opinion `Animal` query would look like the following:


```
{
    Animal {
        name,
        animal_details {
            color
        },
        animal_registration {
            status
        }
    }
}
```

ignasposka (approved):
Good to go, but please fix animal query after merging this.

src/sql-queries/animal.ts
```diff
@@ -40,17 +27,6 @@ export const getAnimalsQuery = (): QueryConfig => {
                     registration_date,
```
psaulenas, line 27: You missed to remove `registration_no` and `registration_date`

psaulenas (approved):
Overall looks good, you missed to remove few columns from one query

PR #44, opened 2021-01-08
CORS policy

tautvydasr:
Add CORS policy without it GQL front client requests will be blocked.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1176,14 +1176,12 @@
     "@sindresorhus/is": {
       "version": "0.14.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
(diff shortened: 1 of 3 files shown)


AzisK (approved):
LG

psaulenas (approved):

ignasposka (changes requested):
There are some doubts about CORS. @JBK91 showed that allowing CORS could result into performance issues: ttps://www.freecodecamp.org/news/the-terrible-performance-cost-of-cors-api-on-the-single-page-application-spa-6fcf71e50147/ and https://blog.bitsrc.io/how-and-why-you-should-avoid-cors-in-single-page-apps-db25452ad2f8.
We can use api.petbook.lt subomain for an api, would that help to avoid allowing CORS?

ignasposka (approved):
Alright, so I think in slack chat we have decided that CORS policy is not that performance killer in our case, as our API has a single endpoint and it requires single OPTIONS request which can be cached by a browser for a relatively long time.

PR #47, opened 2021-01-11
38 - Add fetch query to get available genders

dainius-jankunas:


ignasposka (approved):
Looks good!

src/schemas/gender.ts
```diff
@@ -0,0 +1,35 @@
...
+import getGendersQuery from '../sql-queries/gender';
+
+const typeDef = `
+extend type Query {
+  """
+    Get all genders.
+  
+    Examples:
+  
+    genders(language: 'lt')
```
AzisK, line 11: Single quoted `'lt'` gives me an error. It is only happy with double quotes `"lt"`
```
{
  genders(language: "lt") {
    id
  }
}
```

src/sql-queries/gender.ts
```diff
@@ -0,0 +1,19 @@
+import { QueryConfig } from 'pg';
+
+const getGendersQuery = (language: string): QueryConfig => {
+    const text = `SELECT 
+                    gender as id, 
+                    translation as value
+                FROM gender_translation
+                WHERE gender = ANY(enum_range(null::gender))
```
AzisK, line 8: Do we need this WHERE? Enum itself allows only enum values inside, so this additional check is not needed, I would say

AzisK (changes requested):
Looking good overall. Left a few comments for improvements

src/schemas/gender.ts
dainius-jankunas: fixing that

src/sql-queries/gender.ts
dainius-jankunas: will remove it

JustasBelevicius (approved):

src/sql-queries/gender.ts
```diff
@@ -0,0 +1,19 @@
+import { QueryConfig } from 'pg';
+
+const getGendersQuery = (language: string): QueryConfig => {
+    const text = `SELECT 
+                    gender as id, 
+                    translation as value
+                FROM gender_translation
+                WHERE language = $1
+                ORDER BY id;`;
```
AzisK, line 9: We did not order Animals, so I suppose we should not order Genders as well

AzisK (approved):
I suppose we should not order genders since we did not order animals. Everything else looks smooth

PR #50, opened 2021-01-12
updated animal entity

JBK91:


src/schemas/animal_details.ts
```diff
@@ -0,0 +1,75 @@
...
+  
+    Examples:
+  
+    animals_details
+  """
+  animals_details : [AnimalDetails]
+  }
+  
+"Represents an animal details."
+type AnimalDetails {
```
AzisK, line 28: Originally `animal` table in the database had a field called `status` that was meant to describe health of the animal. Now the `animal` still has it but I am wondering where would it be better to have the health 'status'. In animal directly or in animal_details. WDYT?

AzisK (approved):
Great job

src/schemas/animal_details.ts
AzisK: I also remembered that we should also have status `adopted`. So this could be a separate field or this value could be part of status as well

JBK91: IMO `health` and `adopted` or `not adopted` describes different status, so should be separate field

AzisK: Okay, we can deal with this a little bit later

JustasBelevicius (approved):

ignasposka:
I would suggest use camelCase for naming files in TS. https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#filename

src/sql-queries/animal_details.ts
```diff
@@ -0,0 +1,43 @@
+import { QueryConfig } from 'pg';
+
+export const getAnimalDetailsQuery = (id: number): QueryConfig => {
+    const text = `SELECT
```
ignasposka, line 4: I know it was taken from pg client docs, but maybe it would make more sense to call it somehow different than just "text"? Maybe `sqlQuery`?

src/schemas/animal_details.ts
```diff
@@ -0,0 +1,75 @@
...
+  "Animal id, for example 2"
+  animal_id: Int!
+  "Breed id"
+  breed_id: Int
```
ignasposka, line 32: It should be breed object containing breed_id and translation. In SQL level you should make a join with `breed_translation` and get translations. This is needed for frontend to receive both id and display value when fetching animal. You can see example of how it's done in https://github.com/pets-oss/pets-back/blob/main/src/schemas/gender.ts

src/schemas/animal_details.ts
```diff
@@ -0,0 +1,75 @@
...
+  "Animal name"
+  name: String
+  "Animal gender"
+  gender: String,
```
ignasposka, line 36: Same for gender, you can just reuse https://github.com/pets-oss/pets-back/blob/main/src/schemas/gender.ts

src/schemas/animal_details.ts
```diff
@@ -0,0 +1,75 @@
...
+  "Animal color id"
+  color: Int
```
ignasposka, line 38: Same for colors. Of course breed query is not written yet. It will be written in this ticket https://github.com/pets-oss/pets-back/projects/1#card-52087873
So it's up to you write now or add it later.

ignasposka (changes requested):
For all data fields which have translations it should be added object containing id and translation. For example:
```
{
    "id": "1",
    "text": "Patinas" 
}
```

For getting gender query is already written https://github.com/pets-oss/pets-back/blob/main/src/schemas/gender.ts maybe it can be reused.

src/sql-queries/animal_details.ts
AzisK: We have "text" for AnimalQuery, so I believe @JBK91 is staying consistent. We can have this renaming improvement with a separate PR

ignasposka: Alright, let's leave it like this for now.

src/schemas/animal.ts
```diff
@@ -64,6 +73,20 @@ const resolvers: IResolvers = {
       return dbResponse.rows[0];
     },
   },
+  Animal: {
+    details: async ({ id }, __, { pgClient }) => {
```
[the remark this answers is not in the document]
JBK91: That for animal table and it has `details: AnimalDetails` this query is to get AnimalDetails from animal_details table and it has different params. It takes animal entity as first param, query in `animal_details.ts` takes id from second param from request

src/schemas/animal_details.ts
JBK91: as there is no queries for breed and gender and there is different task for it, and it should be done with other task IMO

src/schemas/animal_details.ts
JBK91: For ideal we should started task from that queries which don't depend on other and then take others when all depended queries exists. But at this moment IMO we should add dependencies in existing queries when writing these queries in separate tasks, and not in one PR, as animal entity is largest

ignasposka (approved):
It turns out we cannot reuse "get all available genders" query in animal entity. Therefore I agree that these fields detalization should be done in another PR.

PR #51, opened 2021-01-12
Add testing solution and write first test

AzisK:
I ended up writing integration tests fetching data from the local server and database.

How did I go with testing our GraphQL endpoint? First of all, I test whether the response is 200, of course. Then I test (validate) that the response (of Animal) has all the expected fields as well as the correct types. Then I do pretty much the same for first Animal of Animals but additionally check if Animals is an array and whether the length of Animals is 5. Checking the length might not be needed but I think it could be useful when changing schemas and data in Postgresql. If there is an error somewhere in the process, we might not have all of the animals inserted.

What inspired me to go with such an approach? I wrote the initial code based on this article https://kolosek.com/testing-graphql-server/. Then I changed the code to Typescript looking at this https://dev.to/daniel_werner/testing-typescript-with-mocha-and-chai-5cl8 and this https://journal.artfuldev.com/write-tests-for-typescript-projects-with-mocha-and-chai-in-typescript-86e053bdb2b6. Then I improved the testing (validating) fetched data part according to this article https://blog.logrocket.com/dynamic-type-validation-in-typescript/. It adds some additional complexity but I quite liked this approach.

Some additional information can be found in updated README.

This is the first first time I write Typescript, test GraphQL and use any of these libraries, so I am open for suggestions.


test/animal.graphql.test.ts
```diff
@@ -0,0 +1,49 @@
+import chai from 'chai';
+import supertest from 'supertest';
+import validate from './animal.interface.validator';
+
+const { expect } = chai;
+const url = `http://localhost:8081`;
```
JBK91, line 6: IMO url should be saved in properties or env file

JBK91 (approved):

test/animal.graphql.test.ts
AzisK: Moved to test/.env

Issue #29, opened 2020-12-23
Set up database update during deploy


ignasposka:
New sql files has to be picked up and applied to database during deploy operation.

PR #52, opened 2021-01-13
#29 - Create database deploy action

JustasBelevicius:


.github/workflows/nuke_database.sql
```diff
@@ -0,0 +1,23 @@
+DO $$
```
JustasBelevicius, line 1: We wouldn't need this if our SQL files contained `DROP IF EXISTS` logic before the creation.

.github/workflows/deploy-database.yml
```diff
@@ -0,0 +1,22 @@
+name: Nuke & Redeploy Database (not prod safe)
```
ignasposka, line 1: Can we make it prod safe initially?

ignasposka (approved):

.github/workflows/deploy-database.yml
AzisK: I believe @JustasBelevicius has a plan for it to create a separate folder just updating and not nuking in the future. For development stage until MVP, I believe we are good to go with this

AzisK (approved):
Looks alrighty

PR #53, opened 2021-01-14
39 - Add fetch query to get available statuses

dainius-jankunas:


test/status.graphql.test.ts
```diff
@@ -0,0 +1,30 @@
+import chai from 'chai';
+import supertest from 'supertest';
+
+require('dotenv').config({ path: './test/.env' });
+
+const { expect } = chai;
+const url = process.env.TEST_URL || 'http://localhost:8081';
+const request = supertest(url);
+
+describe('GraphQL integration tests', () => {
```
ignasposka, line 10: I would suggest to call this "Getting statuses" or something more specific.

test/status.graphql.test.ts
```diff
@@ -0,0 +1,30 @@
...
+          value
+        }}
+      `
+            })
+            .expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
+                const { body: { data: { statuses } } } = res;
+                expect(statuses).to.be.an('array');
+                expect(statuses).to.have.lengthOf(4);
```
ignasposka, line 26: Why don't you validate contents of result? I think we should check that.

ignasposka (changes requested):
Good job, but I think contents of the result should be validated in test.

database/1-schema.sql
```diff
@@ -36,6 +36,22 @@ CREATE TRIGGER app_user_mod_time
...
+CREATE TABLE status_translation (
+    status status,
+    language VARCHAR(4),
+    translation VARCHAR(20),
+    PRIMARY KEY (status, language)
+);
+
+COMMENT ON COLUMN status_translation.language is 'Language code based on BCP 47';
+
+INSERT INTO status_translation (status, language, translation) VALUES
```
AzisK, line 49: I believe it's good to separate insertions to the 2-data file because it's easier to understand the schema when there are only table creations in this file

test/status.graphql.test.ts
AzisK: Yes, there is an example (animal) of how to do that easily. Also "Test" part "C.i. Validate data with interfaces" in README should help you to understand the context and how to validate with interface

test/status.graphql.test.ts
```diff
@@ -0,0 +1,30 @@
...
+const { expect } = chai;
+const url = process.env.TEST_URL || 'http://localhost:8081';
+const request = supertest(url);
+
+describe('GraphQL integration tests', () => {
+    it('Returns all statuses', (done) => {
+        request.post('/graphql')
+            .send({
+                query: `
+        {statuses(language: "lt") {
```
AzisK, line 15: Minor, but the alignment makes it a little hard to read

AzisK (approved):
Good and fast job, left a few comments about possible improvements

AzisK (changes requested):
Agree that we should validate the content of GraphQL

database/1-schema.sql
```diff
@@ -36,6 +36,22 @@ CREATE TRIGGER app_user_mod_time
     EXECUTE PROCEDURE moddatetime (mod_time);
 
 CREATE TYPE status AS ENUM ('healthy', 'vaccinated', 'sick', 'adopted');
+
+CREATE TABLE status_translation (
+    status status,
```
JBK91, line 41: IMO status, language and translation should have constraint NOT NULL. There is no benefit for having NULL in any of these fields

AzisK (approved):

ignasposka (approved):

PR #54, opened 2021-01-14
moved all sql to schema and data files

JBK91:


ignasposka (approved):
Nice cleanup, thanks!

database/2-data.sql
```diff
@@ -1,53 +1,1074 @@
...
+(67, 'lt', 'Gelsvai ruda'),
+(68, 'lt', 'Juoda'),
+(69, 'lt', 'Kreminė'),
+(70, 'lt', 'Pilka'),
+(71, 'lt', 'Ruda'),
+(72, 'lt', 'Rudai raudona');
+
+-- ORGANIZATION
+
+INSERT INTO organization
```
JustasBelevicius, line 995: `organization`, `app_user`, `app_user_roles`, `animal`(and all animal details, and events) data is mostly temporary, so we might want to move it to another file 3-fake-data.sql.

Everything else should stay here since it is data filled from real sources.

JustasBelevicius (approved):
Good job, I left one comment as food for thought.

database/1-schema.sql
```diff
@@ -1,110 +1,174 @@
+-- SPECIES
```
AzisK, line 1: For a better organization I suggest such structure -> organization, app_user, animal, animal related tables, events. Or at least move animal related tables after animal table

database/2-data.sql
```diff
@@ -1,53 +1,1074 @@
...
-VALUES
-(1, '123Svx', 'vaccinated', 'www.url.lt', '2020-08-01', 'Haskelis', 'dog', 'siberian husky', 'female', 'white', 20, '123', '2020-08-11'),
-(2, '321Hi', 'sick', 'www.url.lt', '2020-01-01', 'Jupyteris', 'parrot', NULL, 'female', 'green', 1, '666666', '2020-09-01'),
-(3, '456Carl', 'healthy', 'www.url.lt', '2020-01-01', 'Pitoncas', 'snake', NULL, 'male', 'green', 0.3, '2893402', '2020-03-01'),
-(3, '555Bob', 'adopted', 'www.url.lt', '2020-01-01', 'Javainis', 'hamster', NULL, 'male', 'orange', 0.4, '29387', '2020-04-14'),
-(1, '999Mamba', 'healthy', 'www.url.lt', '2019-09-16', 'Murkė', 'cat', 'ragdoll', 'male', 'grey', 7, '001010101', '2020-09-01');
-
-UPDATE animal SET image_url = 'www.green.mamba' WHERE id = 1;
-
-INSERT INTO animal_event_check_in (animal, date_time, comments, organization)
```
AzisK, line 30: Definitely not a blocker but could we move some of these to events table since they are already here? It might be good to test things later

src/schemas/animal.ts
```diff
@@ -33,6 +33,8 @@ type Animal {
   id: Int!
   "Organization id"
   organization: String!
+  "Animal name"
```
AzisK, line 36: Can you also update the test for animals?

AzisK (changes requested):
Everything is good but I suggest to update animal test and left a few comments for possible improvements

database/2-data.sql
AzisK: I quite like this idea

database/1-schema.sql
JBK91: animal related tables is after animal table. And for animal table and related tables I first need create species, color, gender and etc tables as it is used in animal table as type or reference

database/2-data.sql
JBK91: These tables was removed in other file

src/schemas/animal.ts
JBK91: I did't mention, but I will do it with next PR, as there will be more files with it (need 3 additional interfaces for details, registration and microchip)

database/1-schema.sql
AzisK: Okay, that's fair. However, I still believe that species is part of the animal kingdom, then maybe above animal

database/2-data.sql
AzisK: I am fine for removal, I was talking about data insertions into it - they could be moved to the general events table insertions. However, if it was already deleted, let it be then

src/schemas/animal.ts
AzisK: Okay, that's fine

AzisK (approved):

PR #59, opened 2021-01-15
Create pull-request action

JustasBelevicius:


AzisK (approved):
Lookin' great

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: actions/checkout@v2
+      - name: Setting up node_module cache
+        id: node-modules
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ hashFiles('**/package.json') }}-node-modules
+      - name: Installing node modules...
+        if: steps.node-modules.outputs.cache-hit != 'true'
+        run: npm install
```
AzisK, line 43: Does docker already install npm? If yes, maybe it would be good to move the lint command to be called inside docker pets-back_gis-backend_1 container?

AzisK (approved):
Does docker already install npm? If yes, maybe it would be good to move the lint command to be called inside docker pets-back_gis-backend_1 container?

.github/workflows/pull-request.yml
JustasBelevicius: No reason to, we have unlimited resources AFAIK, and this way we can run the jobs in parallel.

Issue #20, opened 2020-12-15
Fix build warnings


ignasposka:
Fix build warnings:
`npm WARN wcc_gis_backend@1.0.0 No repository field.
npm WARN wcc_gis_backend@1.0.0 No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})`

AzisK:
Also husky@4.3.6 is throwing a silent failure

PR #63, opened 2021-01-16
Fix build warnings #20

AzisK:


psaulenas (approved):

ignasposka (approved):
All good, but as you changing package.json, can you move `nodemon` to devDependencies as it's not needed in prod env?

Issue #45, opened 2021-01-11
Update types


ignasposka:
Types need to be updated according to https://www.vic.lt/gpsas-apskaita/gar-klasifikatoriai/

PR #66, opened 2021-01-17
Update types #45

AzisK:
- Added data from https://www.vic.lt/gpsas-apskaita/gar-klasifikatoriai/
- Added tables breed_category, disease, color_pattern, chip_company_code, install_place
- Updated others
- color_pattern code is not unique (two 04 values), so I had to make primary key out of code and species

Seems weird to add chip_company translations but we have companies as well as words (other, unknown) there.

I did not focus on connecting tables as this is a real data ingestion task. Left connecting all of these tables for another issue. Connected only a few of them, not all.

Additionally
- Renamed chip_install_date to install_date for shorter name


JustasBelevicius (approved):

dainius-jankunas (approved):

database/2-data.sql
```diff
@@ -675,7 +694,7 @@ INSERT INTO breed_translation
 (148, 'lt', 'Lygiaplaukiai retriveriai'),
 (80, 'lt', 'Maljorkos buldogai'),
 (216, 'lt', 'Maltos bišonai'),
-(219, 'lt', 'Mančesterio terjerai'),
+(219, 'lt', 'Mančesterio terjera'),
```
AzisK, line 697: I should fix to `terjerai`

database/2-data.sql
```diff
@@ -749,7 +768,7 @@ INSERT INTO breed_translation
 (275, 'lt', 'Saliukiai'),
 (273, 'lt', 'Samojedai'),
 (290, 'lt', 'San Migelio bandganiai'),
-(274, 'lt', 'Sarloso vilkšuniai'),
+(274, 'lt', 'Sarloso vikšuniai'),
```
AzisK, line 771: I should fix to `vilkšuniai`

ignasposka: Please don't say you wrote everything by hand.

database/2-data.sql
```diff
@@ -765,14 +784,14 @@ INSERT INTO breed_translation
 (296, 'lt', 'Suomių skalikai'),
 (288, 'lt', 'Suomių-lapių špicai'),
 (279, 'lt', 'Šapendojai'),
-(303, 'lt', 'Šar Planinos aviganiai'),
+(303, 'lt', 'Šar Planinos aviganiaiŠ'),
```
ignasposka, line 787: "Š" in the end.

database/2-data.sql
```diff
@@ -765,14 +784,14 @@ INSERT INTO breed_translation
...
 (292, 'lt', 'Šarpėjai'),
 (295, 'lt', 'Šelčiai'),
 (297, 'lt', 'Ši cu'),
-(282, 'lt', 'Šibos'),
-(304, 'lt', 'Šikoku'),
+(282, 'lt', 'Šiba inu '),
+(304, 'lt', 'Šikoku '),
 (281, 'lt', 'Šilerio skalikai'),
-(307, 'lt', 'Šiperkai'),
+(307, 'lt', 'ŠiperkaiŠ'),
```
ignasposka, line 794: "Š" in the end.

ignasposka (approved):
👌

database/2-data.sql
AzisK: I have not although it required a lot of manual work. It was copy paste from the PDF and then superb Sublime multiline functionality and regex magic. Those regex were made on the go and differed on documents. Thus, they were not perfect and we can see some mistakes

PR #67, opened 2021-01-18
35 - Add fetch query to get available breeds

dainius-jankunas:


test/breed.graphql.test.ts
```diff
@@ -0,0 +1,30 @@
...
+            .send({
+                query: `{ breeds(language: "lt", species: "2") {id, code, value} }`
+            })
+            .expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
+                const {body: {data: {breeds}}} = res;
+                expect(breeds).to.be.an('array');
+                validate(breeds[0]);
+                expect(breeds[0].id).to.be.eq('360');
```
ignasposka, line 23: This test would fail in case order of breeds is changed. Maybe there is a way to assert if breeds array have such id at all, instead of checking first element?

ignasposka (approved):

test/breed.graphql.test.ts
AzisK: Atm I am not even sure if we should test such things. This is already not so much GraphQL testing but database schema / insertion testing which is also important and very useful but maybe we should not go into such details

PR #73, opened 2021-01-23
changed field names to camelCase

JBK91:
Updated GraphQL types - changed field names to camelCase
Added field resolver to resolve field: accepts camelCase field, change that field to snake_case and returns that field value

Option for: "It may be good to still have option for us to be able ask for fields in snake_case for debugging reasons" is not done, because GraphQL returns same field names as asked, that means what fields are in Type, so to support camelCase and snake_case we would need 2 same fields in one Type.


AzisK (approved):
GJ

ignasposka (approved):

PR #78, opened 2021-01-26
add authorization to api

povkas:
Will need to add environment variables to the readme


src/index.ts
```diff
@@ -2,6 +2,8 @@ import express from 'express';
 import bodyParser from 'body-parser';
 import cors from 'cors';
 import { graphiqlExpress, graphqlExpress } from 'apollo-server-express';
+var jwt = require('express-jwt');
```
JBK91, line 5: use ES6 syntax (let, const instead of var). Also imports also could be ES6. In this case:
`import { jwt } from 'express-jwt';`

src/index.ts
```diff
@@ -16,9 +18,22 @@ initClients().then(({ pgClient }) => {
     }
   );
 
+  var jwtCheck = jwt({
+    secret: jwks.expressJwtSecret({
+      cache: true,
+      rateLimit: true,
+      jwksRequestsPerMinute: 5,
+      jwksUri: process.env.AUTH0_DOMAIN + '.well-known/jwks.json'
```
JBK91, line 26: Could use template literal. ``${process.env.AUTH0_DOMAIN}.well-known/jwks.json``

JBK91 (changes requested):
Tests are red

ignasposka:
> Will need to add environment variables to the readme

Maybe it is good idea to disable auth for local/dev env?

src/getBearerToken.ts
```diff
@@ -0,0 +1,23 @@
+// axios here is only for testing purposes
```
JBK91, line 1: do we need 2 identical files? `getBearerToken.ts` and `src/getBearerToken.ts `?

JBK91 (approved):

getBearerToken.ts
```diff
@@ -0,0 +1,23 @@
+// axios here is only for testing purposes
+// eslint-disable-next-line import/no-extraneous-dependencies
+import axios from 'axios';
+
+async function getBearerToken() {
+    const data = {
+        client_id: process.argv[3],
```
AzisK, line 7: Would be clearer to access named arguments

AzisK (approved):

src/getBearerToken.ts
ignasposka: Good catch. One file should be eliminated.

PR #79, opened 2021-01-31
68 organization query and mutation

ignasposka:
Added validation solution as well.

### DIFF
--- a/src/sql-queries/organization.ts
+++ b/src/sql-queries/organization.ts
@@ -1,0 +1,56 @@
+import { QueryConfig } from 'pg';
+import { select, insert, update } from 'sql-bricks-postgres';
+import snakeCaseKeys from 'snakecase-keys';
+
+interface CreateOrganisationInput {
+  name: String;
+  country: String;
+  city: String;
+  streetAddress: String;
+  phone: String;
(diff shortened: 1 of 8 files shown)


src/sql-queries/organization.ts
```diff
@@ -0,0 +1,56 @@
...
+  id: number;
+  name: String;
+  country: String;
+  city: String;
+  streetAddress: String;
+  phone: String;
+}
+
+export const getOrganizationQuery = (id: number): QueryConfig =>
+  select().from('organization').where({ id }).toParams();
```
JBK91, line 23: Shouldn't select queries filter rows where delete_time IS NOT NULL?

AzisK, line 23: Should we go with this ORM approach? I believe I would prefer pure SQL as we did before

ignasposka: I guess you're right.

ignasposka: This is not ORM, just sql helper. Partial update gets very messy with pure sql (discussed with @psaulenas).

AzisK (approved):

PR #82, opened 2021-02-01
81 move gql typedefs to separate files

ignasposka:
A lot of new files, because I just moved them to other directories.


AzisK (approved):
LG, did not check very detailed

psaulenas (approved):

PR #84, opened 2021-02-05
77 - add mutation for animal microchip

dainius-jankunas:
no tests yet

### DIFF
--- a/src/schema/typeDefs/animalMicrochip.graphql
+++ b/src/schema/typeDefs/animalMicrochip.graphql
@@ -19,39 +19,39 @@ type AnimalMicrochip {
 
 extend type Mutation {
   "Created microchip"
-  createMicrochip(input: CreateImplantedAnimalMicrochipInput!): AnimalMicrochip
+  createMicrochip(input: CreateAnimalMicrochipInput!): AnimalMicrochip
   "Updated microchip"
-  updateMicrochip(input: UpdateImplantedAnimalMicrochipInput!): AnimalMicrochip
+  updateMicrochip(input: UpdateAnimalMicrochipInput!): AnimalMicrochip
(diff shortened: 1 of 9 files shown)


src/schema/typeDefs/animalMicrochip.graphql
```diff
@@ -7,9 +7,53 @@ type AnimalMicrochip {
...
+extend type Mutation {
+  "Created microchip"
+  createMicrochip(input: CreateImplantedAnimalMicrochipInput!): AnimalMicrochip
+  "Updated microchip"
+  updateMicrochip(input: UpdateImplantedAnimalMicrochipInput!): AnimalMicrochip
+  "Deleted microchip"
+  deleteMicrochip(animal_id: Int!, microchip_id: String!): AnimalMicrochip
+}
+
+input CreateImplantedAnimalMicrochipInput {
```
AzisK, line 31: Maybe no need for `UpdateImplantedAnimalMicrochipInput` and `CreateImplantedAnimalMicrochipInput` since they are the same?

AzisK (approved):
Looks good, left one comment. It would be good to add tests as well

src/schema/typeDefs/animalMicrochip.graphql
ignasposka: Create input object should come without id, because id is being selected by API for a new object.

ignasposka (changes requested):
Good job, but please add tests.

src/schema/typeDefs/animalMicrochip.graphql
JBK91: What cases are where would need change id manually?

ignasposka (approved):
👌

PR #85, opened 2021-02-06
Mutations for animal registration

AzisK:


test/animalRegistration.graphql.test.ts
```diff
@@ -0,0 +1,162 @@
...
+require('dotenv').config({ path: './test/.env' });
+
+const url = process.env.TEST_URL || 'http://localhost:8081';
+const request = supertest(url);
+
+describe ('animalRegistration Graphql mutations tests', () => {
+    it ('Create animal_id=6 registration', (done) => {
+        const mutation = 'createAnimalRegistration'
+        const create = `{
+            animalId: 6,
```
ignasposka, line 14: We shouldn't let choose id for API user, because ids are just auto-incremented.

test/animalRegistration.graphql.test.ts
```diff
@@ -0,0 +1,162 @@
...
+            .expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
+                expect(JSON.stringify(res.body.data[mutation]))
+                .equal(JSON.stringify(answer))
+                return done();
+            });
+    });
+
+    it ('Update animal_id=3 registration', (done) => {
```
ignasposka, line 43: Maybe you can add option when not all fields are present?

ignasposka (changes requested):
I think some changes are needed.

src/sql-queries/animalRegistration.ts
```diff
@@ -18,4 +41,51 @@ const getActiveAnimalRegistrationQuery = (id: number): QueryConfig => {
...
+        values: [
+            snakeInput.animal_id,
+            snakeInput.registration_no,
+        ],
+    };
+
+    return query;
+};
+
+export const undeleteAnimalRegistrationQuery = (input: AnimalRegistrationInput): QueryConfig =>
```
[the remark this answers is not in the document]
AzisK: I imagined a case when we delete a `registrationNo` for an animal and yet later want to add it. It would be impossible to do so because of the primary key restriction. Then this `undeleteAnimalRegistrationQuery` is used. It is also not exposed to API, this is only internal as part of `createAnimalRegistration`

src/schema/resolvers/animalRegistration.ts
```diff
@@ -0,0 +1,42 @@
...
+  Query: {
+    registration: async (_, { id }, { pgClient }) => {
+      const dbResponse = await pgClient.query(getActiveAnimalRegistrationQuery(id));
+      return dbResponse.rows[0];
+    },
+  },
+  Mutation: {
+    createAnimalRegistration: async (_, { input }, { pgClient }) => {
+      const isAnimalRegistration = await pgClient.query(isAnimalRegistrationQuery(input));
+      if (isAnimalRegistration.rows[0] && isAnimalRegistration.rows[0].delete_time) {
```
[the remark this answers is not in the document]
AzisK: It checks whether the registration is deleted. If yes, it undeletes it. Otherwise it creates a new one

test/animalRegistration.graphql.test.ts
AzisK: I think it's fine to have here since the user can know the animal id. How else can we create a registration for a specific animal id without passing it?

test/animalRegistration.graphql.test.ts
AzisK: I could add such a case though I want to say that we should be thoughtful when creating tests because tests have a maintenance price as well

database/1-schema.sql
```diff
@@ -164,8 +164,10 @@ CREATE TYPE registration_status AS ENUM ('Active', 'Inactive');
 CREATE TABLE animal_registration (
     animal_id INTEGER REFERENCES animal(id) NOT NULL,
     registration_no VARCHAR(256) NOT NULL,
-    registration_date DATE,
+    registration_date DATE DEFAULT CURRENT_DATE,
```
JBK91, line 167: registration date is when animal had registration, not when we added in our database

AzisK: I understand that and I believe it should be sent together but imagine a case when the API user forgets to send the registration_date. Then at least this default mechanism sets it to the day it is added

JBK91: But if API user forgets send animal name we don't set it to default name. Also registration_date can't be null, so there should be validation in frontend

src/schema/resolvers/animalRegistration.ts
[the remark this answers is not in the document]
ignasposka: I don't think we should let bring back to life deleted objects so easily. You should inform user that object is deleted and if user really wants to undelete it, he/she will do that action explicitly.

test/animalRegistration.graphql.test.ts
ignasposka: Sorry, I missed the point here.

test/animalRegistration.graphql.test.ts
ignasposka: Alright, that's up to you to decide then. It was just a suggestion.

database/1-schema.sql
ignasposka: Isn't animal having registered when we add it to our database?

AzisK: I don't know if name is a good comparison but I get the point. We do not have registration date `NOT NULL`. However, we can add it. If we do not have it in the database, I think the system would be more robust if we have this default value. I believe we can let the user decide - either go with the default value or change it to whatever he prefers

ignasposka: Yes, I would agree to set current date when registration date is not specified. It is mostly true, that animal is being registered when record is being added to the system.

AzisK:
Updated according to discussions

Moreover, we are able to have multiple registrations for animal (according to database restrictions). Therefore, I made it to return the last (by registration date) active registration with an animal by default and created a separate query to get all the registrations of an animal (by id)

src/schema/typeDefs/animalRegistration.graphql
```diff
@@ -34,14 +45,27 @@ input AnimalRegistrationInput {
   status: RegistrationStatus
 }
 
+input UpdateAnimalRegistrationInput {
+  "Animal id, for example 2"
+  animalId: Int!
+  "Registration number"
+  registrationNo: String!
```
ignasposka, line 52: I don't quite get why we need to have `newRegistrationNo`. If  `registrationNo` is provided - we update this field, otherwise not.

src/sql-queries/animalRegistration.ts
```diff
@@ -10,19 +10,27 @@ enum AnimalRegistrationStatus {
...
+    status: AnimalRegistrationStatus;
+}
+
 interface AnimalRegistrationDeleteInput {
     animalId: number;
     registrationNo: String;
 }
 
-export const getActiveAnimalRegistrationQuery = (id: number): QueryConfig => {
+export const getActiveLastAnimalRegistrationQuery = (id: number): QueryConfig => {
```
ignasposka, line 33: I don't think we should let have more than one active registration at once for animal.

ignasposka (changes requested):

src/sql-queries/animalRegistration.ts
```diff
@@ -50,55 +35,32 @@ export const getActiveLastAnimalRegistrationQuery = (id: number): QueryConfig =>
...
+    .where({ animal_id: input.animalId })
+    .returning(returnFields)
+    .toParams();
+}
+
+export const deleteAnimalRegistrationQuery = (id: number): QueryConfig => {
+    const text = `DELETE
+                FROM ${table}
+                WHERE animal_id = $1
+                RETURNING
```
ignasposka, line 54: Is it going return just nulls? (Because data is deleted)

test/animalRegistration.graphql.test.ts
```diff
@@ -8,40 +8,24 @@ require('dotenv').config({ path: './test/.env' });
...
-            return done();
-        });
-    })
-})
-
 describe ('animalRegistration Graphql mutations tests', () => {
-    const registrationNo = `2021PandemicC19X${uuidv4()}`
+    const registrationNo = `2021PandemicC19X${uuidv4()}`;
+    const date = "2021-01-01";
+    const dateIntString = new Date("2021-01-01").getTime().toString();
```
ignasposka, line 14: Did you want to name it "In" instead of "Int"?

database/1-schema.sql
```diff
@@ -162,11 +162,11 @@ CREATE TABLE animal_details (
 CREATE TYPE registration_status AS ENUM ('Active', 'Inactive');
 
 CREATE TABLE animal_registration (
-    animal_id INTEGER REFERENCES animal(id) NOT NULL,
+    animal_id INTEGER PRIMARY KEY REFERENCES animal(id),
     registration_no VARCHAR(256) NOT NULL,
```
ignasposka, line 166: Maybe mark this as UNIQUE?

ignasposka (approved):
GJ!

test/animalRegistration.graphql.test.ts
AzisK: I wanted Int. It is the date integer but I turn it to string

src/sql-queries/animalRegistration.ts
AzisK: It returns the deleted animal data

PR #88, opened 2021-02-13
Applied nested mutation for animal

ignasposka:
As nested optional fields are one of the GQL advantages, `animalRegistration` mutation can be done inside animal mutation.
By this we have more integral data and we don't have to treat both entities separately.


src/schema/typeDefs/animalRegistration.graphql
```diff
@@ -29,39 +25,26 @@ input CreateAnimalRegistrationInput {
...
     }) {
-      animalId
       registrationNo
       registrationDate
       status
     }
   } ***
   """
-  updateAnimalRegistration(input: UpdateAnimalRegistrationInput!): AnimalRegistration
+  updateAnimalRegistration(input: AnimalRegistrationInput!): AnimalRegistration
```
AzisK, line 47: registrationNo should not be a necessary field for update

src/sql-queries/animalRegistration.ts
```diff
@@ -38,7 +38,7 @@ export const getActiveAnimalRegistrationQuery = (id: number): QueryConfig => {
 export const createAnimalRegistrationQuery = (input: AnimalRegistrationInput): QueryConfig =>
     insert(table, snakeCaseKeys(input))
     .returning(returnFields)
-    .toParams();
+    .   toParams();
```
AzisK, line 41: Typo

test/animal.graphql.test.ts
```diff
@@ -8,37 +8,51 @@ require('dotenv').config({ path: './test/.env' });
...
-                return done();
-            });
-    });
+                }`,
+      })
+      .expect(200)
+      .end((err, res) => {
+        if (err) {
+          // eslint-disable-next-line no-console
+          console.log(res.body);
```
AzisK, line 24: I believe we should get rid of these

AzisK: They could stay though

test/validators/animal.interface.validator.ts
```diff
@@ -211,11 +202,11 @@ export const isAnimal = ajv.compile(AnimalSchema) as ValidateFunction<Animal>;
 export default function validate(value: unknown): Animal {
   if (isAnimal(value)) {
     return value;
-  } else {
+  } 
     throw new Error(
-      ajv.errorsText(isAnimal.errors!.filter((e: any) => e.keyword !== 'if'), {dataVar: 'Animal'}) +
-      '\n\n' +
-      inspect(value),
+      `${ajv.errorsText(isAnimal.errors!.filter((e: any) => e.keyword !== 'if'), {dataVar: 'Animal'})
```
AzisK, line 207: I suggest not to change these parts by hand

package.json
```diff
@@ -14,6 +14,7 @@
     "dev": "tsc && nodemon dist/src/index.js",
     "debug": "tsc && nodemon --inspect-brk=0.0.0.0 dist/src/index.js",
     "test": "mocha -r ts-node/register test/*.test.ts",
+    "testSpecific": "mocha -r ts-node/register",
```
AzisK, line 17: Maybe "test*" would be shorter

AzisK (changes requested):
Please take care of test for this to pass. Also left some comments.
It's definitely useful having this approach. However, we lose the ability to change / create registration separately from the animal. We need to send the whole animal mutation to just create the registration. I believe it would be good to have this possibility separately as well

ignasposka:
> Please take care of test for this to pass. Also left some comments.
> It's definitely useful having this approach. However, we lose the ability to change / create registration separately from the animal. We need to send the whole animal mutation to just create the registration. I believe it would be good to have this possibility separately as well

I don't think that there is a point of creating animal registration without animal itself. That even wouldn't be possible, because animal registration requires animal id.

test/validators/animal.interface.validator.ts
ignasposka: I just ran generator again on this to update validator.

package.json
ignasposka: Star is being ignored in this case, so I cannot use such name.

AzisK:
> > Please take care of test for this to pass. Also left some comments.
> > It's definitely useful having this approach. However, we lose the ability to change / create registration separately from the animal. We need to send the whole animal mutation to just create the registration. I believe it would be good to have this possibility separately as well
> 
> I don't think that there is a point of creating animal registration without animal itself. That even wouldn't be possible, because animal registration requires animal id.

I meant rather the request JSON (knowing animal id), I'll ask front-end how do they imagine these things

AzisK (approved):

PR #89, opened 2021-02-14
Added User CRUD

JBK91:


src/schema/typeDefs/user.graphql
```diff
@@ -0,0 +1,67 @@
...
+  organizationId: Int!
+  "User role in organization"
+  roleType: String
+}
+
+extend type Mutation {
+  "Created user"
+  createUser(input: UserInput!): User
+  "Updated user"
+  updateUser(id: String, input: UserInput!): User
```
JBK91, line 51: left id as param, will remove

AzisK (approved):

test/user.grapqhl.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+describe ('GraphQL user integration tests', () => {
+    it ('Returns user id="aiubfaw4io09" with all fields', (done) => {
+        request.post('/graphql')
+            .send({
+                query: `{ user(id: "aiubfaw4io09") ${userFields} }`
+            })
+            .expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
```
ignasposka, line 19: I would suggest to console log `res.body` in case of error, by that we could read failure reason from test log.

test/user.grapqhl.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            .send({
+                query: `{ users ${userFields} }`
+            })
+            .expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
+                const { body: { data: { users } } } = res;
+                expect(users).to.be.an('array');
+                validate(users[0]);
+                expect(users).to.have.lengthOf(5);
```
ignasposka, line 36: I would suggest to use `to.have.length.above`, because this expect might cause problems when creating more users locally.

test/user.grapqhl.test.ts
```diff
@@ -0,0 +1,128 @@
...
+                            surname: "TestSurname",
+                            email: "TestEmail"
+                        }) ${userFields}
+                    }`
+            }).expect(200)
+            .end((err, res) => {
+                if (err) return done(err);
+                const { body: { data: { createUser } } } = res;
+                validate(createUser);
+                expect(createUser.id).to.be.equals('TestID')
```
ignasposka, line 59: I think it's better to create "expected answer" object and compare it with result once instead of comparing each value separately. Chai testing framework is capable of showing difference between objects in case of test failure.

test/user.grapqhl.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            });
+    });
+
+    it ('Creates user with all fields', (done) => {
+        request.post('/graphql')
+            .send({
+                query:
+                    `mutation {
+                        createUser(input: {
+                            id: "TestID",
```
ignasposka, line 47: You should use string template and reuse field values both in request composing and result validation.

src/sql-queries/user.ts
```diff
@@ -0,0 +1,37 @@
...
+export const getUserQuery = (id: number): QueryConfig =>
+    select().from('app_user').where({ id,}).toParams();
+
+export const getUsersQuery = (): QueryConfig =>
+    select().from('app_user').toParams();
+
+export const createUserQuery = (input: UserInput): QueryConfig =>
+    insert('app_user', snakeCaseKeys(input))
+        .returning(
+            'id, username, name, surname, email, mod_time'
```
ignasposka, line 22: This string could be moved to variable `returningFields`.

ignasposka (approved):
Good job, but tests could be improved (see my comments).

test/user.grapqhl.test.ts
AzisK: I don't think that tests would need to have this. When fixing tests, we can log locally and there is no need to commit this

PR #96, opened 2021-02-19
75 - Add mutations for animalDetails

dainius-jankunas:


src/schema/resolvers/animal.ts
```diff
@@ -39,12 +43,28 @@ const resolvers: IResolvers = {
...
+
         await pgClient.query('COMMIT');
 
-        return {
+        const result = {
           ...createAnimalResult.rows[0],
           registration: createRegistrationResult.rows[0],
         };
+
+        if (!createDetailsResult) {
```
ignasposka, line 61: I think there is nothing wrong returning empty details object, just use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `createDetailsResult?.rows[0]`

ignasposka (approved):

PR #97, opened 2021-02-20
Renamed env file, because it will be used not only for DB.

ignasposka:


psaulenas (commented):
Maybe would be better to have several environment files for different purposes?

JustasBelevicius:
It probably would be a good idea to add database.env.sample and common.env.sample files with placeholder values, as well as document them in the readme.md

psaulenas (approved):

AzisK (approved):

PR #100, opened 2021-02-24
Added tests for organization CRUD.

ignasposka:


src/schema/typeDefs/organization.graphql
```diff
@@ -35,7 +33,7 @@ type Organization {
   "Phone"
   phone: String
   "Modification time"
-  modTime: String
+  modTime: String!
```
JBK91, line 36: It has mod_time when organization is created, but not modified?

JBK91 (approved):

src/schema/typeDefs/organization.graphql
AzisK: Yes, default mod_time should be the creation time

AzisK (approved):

PR #101, opened 2021-02-24
Added eol-last rule, updated eslintignore.

ignasposka:


.eslintrc.js
```diff
@@ -10,4 +10,7 @@ module.exports = {
   parserOptions: {
     project: './tsconfig.json',
   },
+  "rules": {
+    'eol-last': ["warn", "always"],
+}
```
AzisK, line 15: Not a blocker but this curly brace is not aligned

AzisK (approved):

PR #102, opened 2021-02-26
91 - Apply nested mutation for microchip

dainius-jankunas:


test/animalMIcrochip.graphql.test.ts
```diff
@@ -1,112 +1,87 @@
...
-            installDate: dateIntString,
-            installPlace: 1,
-            status: "Implantuotas"
-        }
+    const registrationNo = `2021PandemicC19X${uuidv4()}`;
+    const date = '2021-01-01';
+    const dateIntString = new Date(date).getTime().toString();
 
-        request.post('/graphql')
+    before((done) => {
```
ignasposka, line 17: You could use `test/helpers/createAnimalHelper.ts` for this test preparation. You just need to add microchip data into that helper.

test/animalMIcrochip.graphql.test.ts
```diff
@@ -1,112 +1,87 @@
 import { expect } from 'chai';
```
ignasposka, line 1: Nit: file name has word microchip with second uppercase letter.

ignasposka (approved):

test/animalMIcrochip.graphql.test.ts
dainius-jankunas: how about side effect when testing animal registration we create microchip record? it will do more than needed, so I think that this method should be more common, and not creating all the nested records like details, microchip (unless we pass optional part for that). Let me know if I am wrong

ignasposka: Well, that's a good insight. Then we could adjust helper to accept parameters `addMicrochip`, `addDetails` and `addRegistration`. Is it would be too much to do with a current PR?

dainius-jankunas: Will do additional param to accept the part of the query instead of 3 booleans, we can refactor this later if needed to configure create animals helper method to fit this needs, but now I think it is enough just to pass the test specific part to this common method. What do you think?

PR #104, opened 2021-02-28
Setup indentation 4

AzisK:
Additionally setup warning for double quotes


src/index.ts
```diff
@@ -1,97 +1,108 @@
 import express from 'express';
 import bodyParser from 'body-parser';
 import cors from 'cors';
-import { graphiqlExpress, graphqlExpress } from 'apollo-server-express';
-import { snakeCase } from 'lodash';
+import {
```
ignasposka, line 4: IMHO this looks much worse than a straight line as before. Maybe we could allow few characters more to preserve such imports inline? Previous line is not that long I think 🤔

src/schema/resolvers/animal.ts
```diff
@@ -1,141 +1,185 @@
...
 const resolvers: IResolvers = {
-  Query: {
-    animals: async (_, __, { pgClient }) => {
-      const dbResponse = await pgClient.query(getAnimalsQuery());
-      return dbResponse.rows;
-    },
-    animal: async (_, { id }, { pgClient }) => {
-      const dbResponse = await pgClient.query(getAnimalQuery(id));
-      return dbResponse.rows[0];
+    Query: {
```
ignasposka, line 30: Same here, looks much less readable when split into multiple lines 😖

ignasposka (changes requested):
Requesting changes, because in some places readability looks much worse at least for me. If you think opposite let's discuss.

src/index.ts
AzisK: I used a VS Code plugin 'Beautify' to auto-reindent all the lines. It has also did this "beauty" according to him. I did not find this important to bring it back to one line. If you think that is important, I can bring it back to the old version

src/schema/resolvers/animal.ts
AzisK: I used a VS Code plugin "Beautify" to auto-reindent all the lines. It has also did this "beauty" according to him. I did not find this important to bring it back to one line. If you think that is important, I can bring it back to the old version

src/index.ts
ignasposka: I think you should adjust `Beautify` settings to allow longer lines, or just bring the old formatting back.

test/user.grapqhl.test.ts
```diff
@@ -23,105 +24,126 @@ describe ('GraphQL user integration tests', () => {
...
             .send({
-                query: `{ users ${userFields} }`
+                query: `{ users ${userFields} }`,
             })
             .expect(200)
             .set('Authorization', `Bearer ${process.env.BEARER_TOKEN}`)
             .end((err, res) => {
                 if (err) return done(err);
-                const { body: { data: { users } } } = res;
+                const {
```
ignasposka, line 37: This multiline split also looks unnecessary.

ignasposka (approved):
Just fix all places such I noted in my comment and you're good to merge.

test/user.grapqhl.test.ts
AzisK: It does not seem possible with prettier, so I will look into other formatters

psaulenas (approved):

PR #114, opened 2021-03-02
34 - add animal delete mutation

### LINKED ISSUE #34: Add delete query for animal entity
Add ability to delete animal.
Mutation should be named "deleteAnimal".

psaulenas:
Closes #34 

Example:
<details>
  <summary>Animal deletion</summary>

```graphql
mutation {
  deleteAnimal(input: {id: 2}) {
    id
    name
  }
}
```
</details>

### DIFF
--- a/src/sql-queries/animal.ts
+++ b/src/sql-queries/animal.ts
@@ -32,6 +32,10 @@ interface UpdateAnimalInput {
     microchip: AnimalMicrochipInput;
 }
 
+interface DeleteAnimalInput {
+    id: number;
+}
+
 export const getAnimalQuery = (id: number): QueryConfig => {
     const text = `SELECT
                     id,
@@ -41,7 +45,7 @@ export const getAnimalQuery = (id: number): QueryConfig => {
                     image_url,
                     comments,
                     mod_time
-                FROM public.animal
+                FROM ${table}
                 WHERE id = $1;`;
 
     const query = {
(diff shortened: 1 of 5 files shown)


src/sql-queries/animal.ts
```diff
@@ -82,3 +86,17 @@ export const updateAnimalQuery = (input: UpdateAnimalInput): QueryConfig => {
     .returning(returnFields)
     .toParams();
 };
+
+export const markAnimalDeletedQuery = (input: DeleteAnimalInput): QueryConfig => {
```
ignasposka, line 90: Why it's called `mark` when it's actually deleting?

ignasposka (approved):

src/sql-queries/animal.ts
```diff
@@ -82,3 +86,17 @@ export const updateAnimalQuery = (input: UpdateAnimalInput): QueryConfig => {
...
+  const text = `DELETE FROM ${table}
+                  WHERE id = $1
+                RETURNING
+                  ${returnFields};`;
```
AzisK, line 94: Should we return all the information when deleting?

AzisK (approved):

src/sql-queries/animal.ts
psaulenas: Why not? Data is only from `animals` table.

PR #116, opened 2021-03-03
fixed animal breed query

JBK91:


AzisK (approved):
Looks good but please add this part in the test as well

PR #120, opened 2021-03-07
Added events fetching

JBK91:


test/interfaces/event.interface.ts
```diff
@@ -0,0 +1,16 @@
+import EventType from './eventType.interface';
+
+enum Category {
+    GENERAL = 'GENERAL',
```
ignasposka, line 4: Why this mapping is needed? Cannot we just use pure value in enum?

test/validators/event.interface.validator.ts
```diff
@@ -0,0 +1,96 @@
+/* tslint:disable */
```
ignasposka, line 1: Add `/* eslint-disable */` as well.

ignasposka (approved):

test/validators/event.interface.validator.ts
AzisK: No need to add `/* eslint-disable */` because `.eslintignore` has `test/validators`. We could actually ignore tslint on validators in general. As in https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

Just add this in `.tsconfig.json`
```
{
    "exclude": ["test/validators"]
}
```

src/sql-queries/event.ts
```diff
@@ -0,0 +1,54 @@
...
+        .toParams();
+
+export const getEventType = (eventId: number, language: string): QueryConfig =>
+    select('event AS id', 'translation AS type')
+        .from('event_translation')
+        .where({ language, event: eventId })
+        .toParams();
+
+export const getAllEvents = (): QueryConfig =>
+    select('*', '\'GENERAL\' AS category')
```
AzisK, line 17: Should we be more specific than `*`?

AzisK (approved):
LG

src/sql-queries/event.ts
JBK91: IMO, no

PR #121, opened 2021-03-07
74 add validation info

ignasposka:
Actual validation in resolvers will be added in another PR.


database/1-schema.sql
```diff
@@ -2,10 +2,10 @@
 
 CREATE TABLE organization (
     id SERIAL PRIMARY KEY,
-    name VARCHAR(256) NOT NULL,
+    name VARCHAR(255) NOT NULL,
```
JustasBelevicius, line 5: 256 is a power of 2, not sure why the change by 1 char?

JustasBelevicius (approved):

database/1-schema.sql
ignasposka: DB for storing input length would use 1 byte in this case. While 256 it would need to consume 2 bytes.

PR #122, opened 2021-03-07
115 update readme

ignasposka:


README.md
```diff
@@ -30,9 +39,15 @@ POSTGRES_HOST=pets-back_db_1
 POSTGRES_PORT=5432
 ```
 
-3. Install and run docker
-4. Run `docker-compose up -d` or `docker-compose up -d --build` if it is not the first time, and the Dockerfile was changed.
-5. Hack away - hot-reloading enabled
+3. Add common.env with the following values
```
JustasBelevicius, line 42: Let's create `database.sample.env` and `common.sample.env` files and rewrite 2. and 3. to something like: "rename the sample env files and adjust values to your liking"

JustasBelevicius (approved):

PR #135, opened 2021-03-22
Break test to check if action runs when PR is raised in fork

JustasBelevicius:


ignasposka:
Can we close it now?

JustasBelevicius:
Fixed

PR #165, opened 2021-04-06
139-Add_cage_field_that_animal_is_placed_into_inside_an_organization

tryDmitriyCatch:


tryDmitriyCatch:
Would be nice if someone looked at this. Thanks!

AzisK (approved):
Good job. I believe that cage field does not need to be indefinitely long, therefore changed it to VARCHAR(128)

Issue #152, opened 2021-04-01
Add mod_time column for animal_microchip table 


AzisK:
**Goal**
Add mod_time column for animal_microchip table

**Description**
Also trigger moddatetime extension before each update of row in that table

Example:
```sql
CREATE TRIGGER table_mod_time BEFORE UPDATE ON table
FOR EACH ROW EXECUTE PROCEDURE moddatetime (mod_time);
```

Current table DDL:
```sql
CREATE TABLE animal_microchip (
    animal_id INTEGER REFERENCES animal(id) ON DELETE CASCADE NOT NULL,
    microchip_id VARCHAR(255) NOT NULL,
    chip_company_code chip_company_code NOT NULL,
    install_date DATE,
    install_place install_place NOT NULL,
    status chip_status DEFAULT 'Implanted',
    PRIMARY KEY (animal_id, microchip_id)
);
```

rimantepomeckyte:
Leave a comment for an assignee

tryDmitriyCatch:
Assign this to me, please!

Issue #141, opened 2021-03-23
Create animal_favorite table to store user's favorite animals


AzisK:
**Goal**

Create animal_favorite table to store user's favorite animals

**Description**

One user can have multiple favorite animals and one animal can be favorite for several users (many to many relationship). Example:

user_id | animal_id
--------|----------
1|1
1|2
2|2

mindra:
https://github.com/pets-oss/pets-back/pull/177 could be solution for this.

rimantepomeckyte:
Please, add me as an assignee

Issue #140, opened 2021-03-23
Create animal_gallery table to store animal gallery images


AzisK:
**Goal**

Create animal_gallery table to store animal gallery images

**Description**

Create animal_gallery table to store animal gallery images where front-end could store and retrieve images from. One animal can have multiple images, e.g.

animal_id | url
-|-
1|url1
1|url2
2|url3

rimantepomeckyte:
Leaving a comment for a asignee

PR #170, opened 2021-04-06
#152, #140, #141 animal gallery, animal favorite, modtime to animal microchip

rimantepomeckyte:


database/1-schema.sql
```diff
@@ -224,6 +225,18 @@ CREATE TABLE former_animal_owner (
     phone VARCHAR(64)
 );
 
+CREATE TABLE animal_gallery (
+    id SERIAL PRIMARY KEY,
+    animal_id INTEGER REFERENCES animal(id) ON DELETE CASCADE NOT NULL,
+    url VARCHAR(255)
```
JustasBelevicius, line 231: Looking at the previous implementation in the animals table, the size used there was 2048, it would be best keeping it consistent.

database/1-schema.sql
```diff
@@ -224,6 +225,18 @@ CREATE TABLE former_animal_owner (
...
+);
+
+CREATE TABLE animal_favorite(
+user_id VARCHAR(255) REFERENCES app_user(id) ON DELETE CASCADE NOT NULL,
+animal_id INTEGER REFERENCES animal(id) ON DELETE CASCADE NOT NULL,
+PRIMARY KEY (user_id, animal_id)
```
JustasBelevicius, line 237: formatting: missing tabs

JustasBelevicius (commented):
Well done! I have left some comments under specific lines.

Besides that, it would also be useful adding fake data for favourite and gallery tables.

P.S. In my opinion, it would've been better to have 3 different commits for each issue (can still be under the same pull request). That way a specific change can be easily reverted if the need arises.

database/1-schema.sql
JustasBelevicius, line 231: Looking at all the issues where we are adding mod_times, we should actually add one here as well. Don't create to add a trigger for it.

JustasBelevicius (changes requested):
Let's focus on the `animal_gallery` functionality with this PR. There are a few improvements you should do before we merge it:

- [ ] Add `mod_time` field and trigger for the `animal_gallery` field
- [ ] Update the size of the `url` to be consistent with other url fields (check `animals` table)
- [ ] Add fake data for this table.

JustasBelevicius (approved):
Looks good, I've selected other PRs for animal favorite and animal microchip, so let's wait until those are merged, and then merge this one in.

.gitignore
```diff
@@ -121,3 +121,5 @@ node_modules
 
 # IntelliJ IDEA
 .idea
+
+package-lock.json
```
AzisK, line 125: From what I know and remember, we should not have `package-lock.json` in `.gitignore`. This file speeds up the deployment process and if new libraries are added, they should be commited. Am I right, @JustasBelevicius?

database/1-schema.sql
```diff
@@ -224,6 +225,19 @@ CREATE TABLE former_animal_owner (
...
 
+CREATE TABLE animal_gallery (
+    id SERIAL PRIMARY KEY,
+    animal_id INTEGER REFERENCES animal(id) ON DELETE CASCADE NOT NULL,
+    url VARCHAR(2048),
+    mod_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+
+CREATE TABLE animal_favorite(
+user_id VARCHAR(255) REFERENCES app_user(id) ON DELETE CASCADE NOT NULL,
```
AzisK, line 236: I know this pull request focuses on animal_gallery but if  animal_favorite differs from main, then conflicts would need to be resolved for animal_favorite table. Thus, it's good to align the spacing here as well

AzisK (changes requested):
Good job in general! I left a few comments for possible improvements

.gitignore
JustasBelevicius: Lets ignore it. If we add new packages we will force push the package-lock.json file, this will prevent others from pushing package-lock.json changes accidentally

database/1-schema.sql
JustasBelevicius: Not a blocker, will be resolved before merging, when resolving conflicts

rimantepomeckyte: So for now i should leave it or fix it?

AzisK: I believe you could leave it then. We can resolve this when resolving the conflict.

rimantepomeckyte: Ok, thanks

AzisK (approved):

PR #171, opened 2021-04-07
142-Add animal_behavior column in animal_details table

Renata-Z:


src/index.ts
```diff
@@ -10,6 +10,7 @@ import initClients from './utils/init-clients';
 import { graphqlUploadExpress } from 'graphql-upload';
 
 const { ApolloServer } = require('apollo-server-express');
+
```
AzisK, line 13: Believe this line is not needed here

AzisK (approved):
Well done. There is one line that should not necessarily be there

src/index.ts
AzisK: On the other hand it makes sense to separate imports from code

AzisK (approved):

Issue #144, opened 2021-03-24
Create a database query that would return statistics of sheltered animals


AzisK:
**Goal**

Create a database query that would return statistics of sheltered animals

**Description**

Create a database query that would return the count of sheltered animals by month in organization. Such events can be found in `animal_event_found` and `animal_event_given_away` tables. Also group it by circumstance and species. Use the Lithuanian translations of species ("lt" language). This grouped data will be later used to generate monthly and yearly statistical reports. Please add this query in file `reports.sql` in the root of repository (can comment above it what is it about). Example table below

year_month | species | circumstance | cnt
-|-|-|-
201801|Katė|found|10
201801|Šuo|found|13
201801|Šeškas|given away|4
201802|Šuo|found|5
201802|Katė|given away|10
201803|Šuo|found|13

PR #174, opened 2021-04-07
144 sheltered animal stats

mindra:
Solving issue https://github.com/pets-oss/pets-back/issues/144
Would be nice to get  feedback


database/4-reports.sql
```diff
@@ -0,0 +1,36 @@
...
+SELECT
+	to_char(aega.date_time, 'yyyymm') AS "year_month",
+	ad.breed_id,
+	a.organization,
+	'given away' AS "circumstance",
+	count(aega.animal_id) AS "cnt"
+FROM animal_event_given_away aega
+INNER JOIN animal_details ad ON ad.animal_id = aega.animal_id
+INNER JOIN animal a ON a.id = ad.animal_id
+GROUP BY to_char(aega.date_time, 'yyyymm'), ad.breed_id, a.organization
```
AzisK, line 23: Good job on even grouping by organization. Pardon me for not specifying this in the ticket. For now we can simply return the sheltered statistics of all organizations and later we can always plug in a WHERE filter to filter by organization id

database/4-reports.sql
```diff
@@ -0,0 +1,36 @@
+-- https://github.com/pets-oss/pets-back/issues/144
+WITH cte_stats AS (
+SELECT
+	to_char(aef.date_time, 'yyyymm') AS "year_month",
+	ad.breed_id,
+	a.organization,
+	'found' AS "circumstance",
+	count(aef.animal_id) AS "cnt"
+FROM animal_event_found aef
+INNER JOIN animal_details ad ON ad.animal_id = aef.animal_id
```
AzisK, line 10: There is no need to write `INNER JOIN` as `JOIN` is defaulted to `INNER JOIN`

database/4-reports.sql
AzisK: According to this ticket, we should group by species, not breed. Thus, I believe grouping by breed_id is not totally correct

database/4-reports.sql
```diff
@@ -0,0 +1,36 @@
...
+)
+SELECT
+	rs.year_month,
+	st."translation" AS "species",
+	rs.circumstance,
+	o.Name AS "Organization",
+	rs.cnt
+FROM cte_stats rs
+INNER JOIN breed b ON b.id = rs.breed_id
+INNER JOIN species_translation st ON st."species" = b."species"
```
AzisK, line 33: As we have translations, we should specify the "lt" language here as well. If not and if we had more translations, we would get all of the translations

AzisK (changes requested):
That's a good start. I left some comments for improvements.

This is a bit shorter solution that I propose (it's possible to come up with an even shorter one)

```sql
SELECT 
	EXTRACT(YEAR FROM aef.date_time) * 100 + EXTRACT(MONTH FROM aef.date_time) AS year_month,
	st."translation", 
	'found' AS circumstance,
	COUNT(*) cnt
FROM animal_event_found aef 
JOIN animal_details ad
    ON aef.animal_id = ad.animal_id 
JOIN breed b 
    ON ad.breed_id = b.id 
JOIN (SELECT * FROM species_translation WHERE language = 'lt') st 
    ON b."species" = st."species" 
GROUP by st."translation", year_month
UNION ALL
SELECT 
    EXTRACT(YEAR FROM agw.date_time) * 100 + EXTRACT(MONTH FROM agw.date_time) AS year_month,
	st."translation", 
	'given away' AS circumstance,
	COUNT(*) cnt
FROM animal_event_given_away agw 
JOIN animal_details ad
    ON agw.animal_id = ad.animal_id 
JOIN breed b 
    ON ad.breed_id = b.id 
JOIN (SELECT * FROM species_translation WHERE language = 'lt') st  
    ON b."species" = st."species" 
GROUP BY st."translation", year_month
ORDER BY year_month;
```

AzisK:
@mindra are you going to update the code?

mindra:
I'll do my best to push updates later today

database/4-reports.sql
mindra: Updated

database/4-reports.sql
mindra: Updated

database/4-reports.sql
mindra: Updated

AzisK (approved):
LGTM

Issue #151, opened 2021-04-01
Normalize address in animal_event_found table


AzisK:
**Goal**
Normalize address in former_animal_owner table

**Description**
Suggestion to normalize is into street address and house number while house number is not a necessary field since sometimes it may be hard to determine next which house the animal was found. Current table structure:

```sql
CREATE TABLE animal_event_found (
    id SERIAL PRIMARY KEY,
    address VARCHAR(256),
    municipality_id INTEGER REFERENCES municipality(id) NOT NULL,
    date_time TIMESTAMP,
    animal_id INTEGER REFERENCES animal(id) ON DELETE CASCADE NOT NULL,
    comments TEXT
);
```

pergaliuke:
I did it! :)

JustasBelevicius:
@ieva1919

PR #176, opened 2021-04-07
#151 normalize address in animal event found table

ieva1919:


JustasBelevicius (commented):
Well done! Not sure what the former_animal_owner address addition is about since the ticket didn't ask for it, but overall the normalisation is done correctly.

Despite that another PR was merged solving the same issue, so I will have to close this one.

Thanks and keep up the good work!

AzisK:
Good job! As @JustasBelevicius said, this issue was done with another PR https://github.com/pets-oss/pets-back/pull/178 already. There I suggested to shorten the column names to simply street and house_no. 

If we were to simply add address to former_animal_owner, that would not be truly normalized.

Keep up the good work!

Issue #145, opened 2021-03-27
Get rid of delete_time in organization table


AzisK:
**Goal** 

Get rid of delete_time in organization table

**Description**

This is a left-over from previous attempt to implement soft-deletes. However, it was abandoned for the MVP version. Make sure this change does not break any tests. Table DDL (data definition language):

```sql
CREATE TABLE organization (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) NOT NULL,
    country VARCHAR(128),
    city VARCHAR(128),
    street_address VARCHAR(255),
    phone VARCHAR(64),
    mod_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
    delete_time TIMESTAMP
);
```

kostamir:
Got rid of delete_time. Fixed the code to pass all the tests.

PR #179, opened 2021-04-08
#145 get rid of delete time in organization table

kostamir:


src/sql-queries/organization.ts
```diff
@@ -1,5 +1,5 @@
 import { QueryConfig } from 'pg';
-import { select, insert, update } from 'sql-bricks-postgres';
+import { select, insert, update, delete as remove } from 'sql-bricks-postgres';
```
AzisK, line 2: Why rename delete to remove?

src/sql-queries/organization.ts
```diff
@@ -21,36 +21,20 @@ interface UpdateOrganizationInput {
...
         .toParams();
 
 export const getOrganizationsQuery = (): QueryConfig =>
-    select()
-        .from('organization')
-        .where({
-            delete_time: null,
-        })
-        .toParams();
+    select().from('organization').where({}).toParams();
```
AzisK, line 31: I believe we can skip the `where` part?

AzisK (changes requested):
Well done. It looks good. Left a comment for a possible improvement and a question

src/sql-queries/organization.ts
kostamir: Done.

src/sql-queries/organization.ts
kostamir: Tried to use delete as is but vscode thew an error so we came up with this workaround.

AzisK (approved):

src/sql-queries/organization.ts
JBK91: Because delete is reserved JS keyword

ignasposka (approved):

Issue #139, opened 2021-03-23
Add cage field that animal is placed into inside an organization


AzisK:
**Goal**

Add cage field that animal is placed into inside an organization

**Description**

This is a field that animal shelters would like to have. Figure out the best database structure for it.

PR #180, opened 2021-04-08
#139 Add cage field that animal is placed into inside an organization

pergaliuke:


AzisK (approved):
Looks good, well done

PR #181, opened 2021-04-08
Issue148 Create a database table organization_task that would store organization tasks

rimantepomeckyte:
Create a database table organization_task that would store organization tasks


psaulenas (approved):
GJ!

rimantepomeckyte:
Leaving a comment for Assignee

Issue #149, opened 2021-04-01
Create a database query that would return current count of animals in organization


AzisK:
**Goal**
Create a database query that would return current count of animals

**Description**
Group this count by species and gender. Use the Lithuanian translations of species and gender ("lt" language). This information of current state will later be used to display the state of the organization or in reports. Please add this query in file `reports.sql` in the root of repository (can comment above it what is it about). For now we can look at the general state of organizations and later (when we have the GraphQL query ready) we can plug-in a WHERE filter to filter out the organization id. Example table below

species | gender | cnt
-|-|-
Šuo|Patelė|2
Katė|Patinas|1
Katė|Patelė|1
Šeškas|Patelė|1

mantasimkus:
As there might be multiple organizations, should the query return:
a) general results among all organizations,
b) results in a specific organization only (using a hardcoded WHERE clause) or
c) results grouped by organization (it probably means storing some additional information in the table, say, organization name)?

AzisK:
a) general results among all organizations
We can always amend the query and add a WHERE filter to filter the organization id

mantasimkus:
I would like this issue to be assigned to me.

PR #182, opened 2021-04-08
#149 Create a database query that would return current count of animals in organization

mantasimkus:
Added a SQL query that counts animals in general grouped by species and gender. If it is required to count animals per organization, the query should be appended with a WHERE clause. Solves issue #149.


reports.sql
```diff
@@ -0,0 +1,16 @@
+-- Current count of animals in organization grouped by species and gender
+
+SELECT st.translation AS species,
+    gt.translation AS gender,
+    COUNT(*) AS cnt
+FROM animal AS a
+    JOIN animal_details AS ad ON a.id = ad.animal_id
+    JOIN breed AS b ON ad.breed_id = b.id
+    JOIN species_translation AS st ON b.species = st.species
+    AND st.language = 'lt'
```
psaulenas, line 10: It would be great if you could extract `lt` value to a variable like in this (example)[https://stackoverflow.com/a/16552441]

reports.sql
```diff
@@ -0,0 +1,25 @@
...
+WITH const AS (
+    SELECT 'lt' AS lang
+)
+SELECT st.translation AS species,
+    gt.translation AS gender,
+    COUNT(*) AS cnt
+FROM animal AS a
+    JOIN animal_details AS ad ON a.id = ad.animal_id
+    JOIN breed AS b ON ad.breed_id = b.id
+    JOIN species_translation AS st ON b.species = st.species
```
AzisK, line 12: The query could have been written this way but both ways work and maybe yours is more readable. I'd move JOIN to the same level as SELECT/FROM commands but that is my personal preference
```sql
WITH const AS (
    SELECT 'lt' AS lang
)
SELECT st.translation AS species,
    gt.translation AS gender,
    COUNT(*) AS cnt
FROM animal AS a
JOIN animal_details AS ad ON a.id = ad.animal_id
JOIN breed AS b ON ad.breed_id = b.id
JOIN (SELECT * from species_translation WHERE language = (SELECT lang FROM const)) st 
    ON b.species = st.species
JOIN (SELECT * from gender_translation WHERE language = (SELECT lang FROM const)) gt 
    ON ad.gender_id = gt.gender
GROUP BY st.translation,
    gt.translation
ORDER BY st.translation,
    gt.translation;
```

JustasBelevicius: I agree with @AzisK here. By moving lang selection inside, and doing join with the result, we eliminate a lot of unnecessary joins. It is hard to see the purpose of such optimisation because we only use LT language right now.

Issue #154, opened 2021-04-01
 Add mod_time column for former_animal_owner table


AzisK:
**Goal**
Add mod_time column for former_animal_owner table

**Description**
Also trigger moddatetime extension before each update of row in that table

Example:
```sql
CREATE TRIGGER table_mod_time BEFORE UPDATE ON table
FOR EACH ROW EXECUTE PROCEDURE moddatetime (mod_time);
```

Current table DDL:
```sql
CREATE TABLE former_animal_owner (
    id SERIAL PRIMARY KEY,
    name VARCHAR(256) NOT NULL,
    surname VARCHAR(256),
    phone VARCHAR(64)
);
```

LukasKri:
Fixed, thank you!

Issue #153, opened 2021-04-01
Fix id column in municipality table


AzisK:
**Goal**
Fix id column in municipality table

**Description**
Right now municiplities are inserted like this

```sql
INSERT INTO municipality (id, name)
VALUES
(1, 'Akmenės rajono savivaldybė'),
(2, 'Alytaus miesto savivaldybė');
```

while table DDL is:

```sql
CREATE TABLE municipality (
    id SERIAL PRIMARY KEY,
    name VARCHAR(256) NOT NULL
);
```

From this we can see that SERIAL is not really needed since we insert id manually. It would also throw an error if we did not insert id because it would start from 1.

Another solution would be to leave SERIAL column but insert without id (let it be handled by SERIAL itself).

Not a big issue because municipalities probably will not change often but it could change at some point. Also this could be useful when we would be implementing multi-language solution for municipalities - not jut Lithuanian names.

Issue #150, opened 2021-04-01
Add mod_time column for animal_details table


AzisK:
**Goal**
Add mod_time column for animal_details table

**Description**
Also trigger moddatetime extension before each update of row in that table

Example:
```sql
CREATE TRIGGER table_mod_time BEFORE UPDATE ON table
FOR EACH ROW EXECUTE PROCEDURE moddatetime (mod_time);
```

Current table DDL:
```sql
CREATE TABLE animal_details (
    animal_id INTEGER PRIMARY KEY REFERENCES animal(id) ON DELETE CASCADE,
    breed_id INTEGER REFERENCES breed(id),
    gender_id gender,
    color_id INTEGER REFERENCES color(code),
    birth_date DATE,
    weight NUMERIC,
    allergy VARCHAR(128),
    food VARCHAR(255)
);
```

ieva1919:
I fixed this issue!

Issue #142, opened 2021-03-23
Add animal_behavior column in animal_details table


AzisK:
**Goal**

Add animal_behavior column in animal_details table

**Description**

This is a field that animal shelters would like to have. Make sure it does not break tests.

pergaliuke:
Assign me too

PR #183, opened 2021-04-08
Database issues A

sunny-github-acc:
#145, #138, #142, #150, #152, #154, #153


AzisK (changes requested):
I see that you addressed a lot of issues in 1 ticket. Good job in general. However, I advise to create pull requests for fewer (ideally 1) issues in the future. It is easier to separate concepts, thus easier to review, less likely to have files strongly deviated from the main branch. Also, because these issues had been introduced as part of learning issues for Academy, I believe it is more fun for everyone to have at least 1 small issue merged (so this point could be different for future issues).

Please make sure that there are no conflicting files and that tests pass if you would like to have this pull request approved

PR #184, opened 2021-04-08
 #153 Fix id column in municipality table, #154 Add mod_time column for former_animal_owner table

rimantepomeckyte:


AzisK (approved):
Well done! Let's focus on id column in municipality table here since the other issue has been done already.

rimantepomeckyte:
Please, add me as an assignee

PR #185, opened 2021-04-10
143 Update Entity Relation Diagram according to the newest schema

LukasKri:
Updated ER diagram (2021-04-09 data) using Lucid, I think I'll be able to update it later on as well.


AzisK:
Impressing. I will look into the details of the diagram later but I must say that I am truly impressed

LukasKri:
Thanks a lot! I'll be waiting for your comments about the possible further improvements.

JustasBelevicius:
Very well done, did you draw the ER diagrams by hand or is there a tool to automate it? It would be interesting to creating a github action which automatically generates an ER diagram.

LukasKri:
I've done it by hand having two windows opened side by side - Lucidchart and VSCode, so there might be some spelling mistakes, although I double checked. I would be quite disappointed in myself, if there is a tool to automate it, lol. I know that you can make a similar diagram using DBeaver, but it would be somewhat incomplete, because DBeaver is not able to indicate foreign keys.

I need to get more familiar with GitHub action functionality in order to find a better solution then.

AzisK:
Yes, DBeaver does not show foreign keys. I see that MySQL Workbench is also able to draw out a schema from a database and it shows foreign keys ([source](https://medium.com/@tushar0618/how-to-create-er-diagram-of-a-database-in-mysql-workbench-209fbf63fd03)). However, your hand-computer drawn ER diagram is still supreme because it shows default values and it shows `ON DELETE CASCADE`. I would also be surprised if MySQL Workbench showed enum values.

Having said that, it is a lot of work to support the database changes in ER diagram manually in the long term. Therefore, automation would be great if there is such a possibility to this. Everything is possible but by possibility I mean not having to build a spaceship. That's an open question to research this. What program let's us do this programmatically? Could we use it in Github Actions? By the way, @LukasKri you would need special permissions to add Github actions to this repo. Also, where would we store this image / PDF? Should the action commit this file to the repository? These are not trivial questions and I believe it would be interesting to address them. This does not have to be done with this issue because it's a rather advanced engineering question. But if this sparks interest, feel free to create such pull request in the future!

Coming back to the ER diagram itself, it is there to actually understand the whole database, inspect it, maybe spot places for improvement, to debug possible issues. I wonder if doing this helped you understand the whole database of this app? Did you spot what is missing or could be improved? I believe this is even more valuable for the app and for the person getting familiar with the app. We will also now have a more recent version of the ER diagram that we could look into. Thanks!

LukasKri:
Thank you, @AzisK for sharing your knowledge and ideas out loud, as I can tell that there is lots of room for improvement. I thought that MySQL Workbench is for MySQL databases only, but I was wrong then. As you say, it may be a better option to use an automatic drawing tool in a long term.

I really enjoyed doing this task, because it let me to get to know the whole database and entity relations better, although I still lack of experience in order to point out the weak spots or offer ways for improvement, my apologies for that.

ignasposka:
@AzisK Can we merge this?

AzisK:
Yes, I believe so

AzisK (approved):
GJ

Issue #118, opened 2021-03-06
Add query to get all available chip companies


ignasposka:
**Goal**

Create a GraphQL query that would return all available chip companies from the database.

**Description**

Front-end needs to be able to retrieve all possible chip companies because it needs to show them for a user when entering animal microchip installation information. Please create such a GraphQL query and write automatic tests that would ensure it works correctly. There are a bunch of examples one can look at. For example, we can look at gender GraphQL definition at https://github.com/pets-oss/pets-back/blob/main/src/schema/typeDefs/gender.graphql. The pull-request that introduced such a query for genders can be seen here https://github.com/pets-oss/pets-back/pull/47 

Thus our chip company GraphQL definition could look like the following

```graphql
extend type Query {
    """
    Get all chipCompanies.

    Examples:

    chipCompanies(language: "lt")
    """
    chipCompanies("language code" language: String!): [ChipCompany]
}

"Represents a chip company."
type ChipCompany {
    "Chip company id"
    id: String!
    "Chip company value"
    value: String!
}
```

thus

**Expected behaviour**

Such GraphQL query

```graphql
{
    chipCompanies(language: "lt") {
        id,
        value
    }
}
```

Should return all chip companies with ids and values in Lithuanian language

PR #187, opened 2021-04-10
#118 Add query to get all available chip companies

mantasimkus:
Created a GraphQL query that returns all chip companies from the database and wrote an automatic test to check if the query works correctly. Solves issue #118.


AzisK (approved):
Looking good, well done!

psaulenas (approved):
GJ

Issue #188, opened 2021-04-12
Expand /status endpoint with more information


katilius:
Status endpoint is usually good way to see essential information about service. Usually this endpoint returns information about service itself state and also sate on each of it's integrations. Because for example even if service is running, but it cannot connect to database, it is not functioning properly.

Current status endpoint only returns OK response and no information.

Expand it to return this info:

- version - current service version. Load this from `package.json` field `version`
- database - check if database connection is ok
- cloudinary - check if cloudinary connection works
- status - ok when database is ok and cloudinary is ok

Example JSON:
```
{
    "status": "ok",
    "database": "ok",
    "cloudinary": "ok",
    "version": "1.0.0"
}
```

PR #190, opened 2021-04-12
#188 expand /status endpoint with more information

ieva1919:


src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
+        }
+
+        try {
+            isVersion = await getVersion()
+        } catch (error) {
+            console.log(error)
+        }
+
+        res.send({
+            'status': 'ok',
```
lukaskairys, line 45: status should only be ok when both database and cloudinary statuses are ok

src/getVersion.ts
```diff
@@ -0,0 +1,7 @@
+import fs from 'fs'
+
+export default async function getVersion() {
+    const result = await fs.promises.readFile('package.json', 'utf8')
```
psaulenas, line 4: You could get the version using import from `package.json`
```js
```

src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
 
 const { ApolloServer } = require('apollo-server-express');
+
 initClients().then(({ pgClient, cloudinaryClient }) => {
     const app = express();
 
-    app.use('/status', (req, res) => {
-        res.sendStatus(200);
+    app.use('/status', async (req, res) => {
+        let isDataBasedActive = false
```
ignasposka, line 19: Typo in word "Database".

src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
+        let isCloudinaryClient = false
```
ignasposka, line 20: It should be called `isCloudinaryClientActive`.

src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
+        let isVersion = ''
```
ignasposka, line 21: It should be called just `version`.

src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
+
+        try {
+            const results = await pgClient.query({
+                text: 'select true as ok'
```
ignasposka, line 25: SQL keywords have convention to be uppercase. Therefore it should be `SELECT true AS ok`.

src/services/cloudinary.ts
```diff
@@ -30,9 +31,13 @@ const cloudinaryClient = (cloudinary: any): CloudinaryClient => ({
     },
     deleteImage: (imageUrl: string) => {
         const parts = imageUrl.split('/');
-        const [ publicId ] = parts[parts.length - 1].split('.');
+        const [publicId] = parts[parts.length - 1].split('.');
         cloudinary.uploader.destroy(publicId);
     },
+    isOk: async () => {
+        const response = await cloudinary.api.ping()
```
ignasposka, line 38: Missing semicolon in the end of a line.

ignasposka (changes requested):

src/index.ts
```diff
@@ -4,17 +4,49 @@ import cors from 'cors';
...
+            })
+            isDataBasedActive = results.rows[0].ok
```
ignasposka, line 27: This access chain could lead to null pointer exception. Use optional chaining to prevent that: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

It would look like `results.rows[0]?.ok`

src/getVersion.ts
ieva1919: I attempted to import this but I have an error Cannot find module '../package.json'. What I have to change for this import to work?

psaulenas: You need to add `resolveJsonModule: true` in the tsconfig file

ieva1919: Thank you, now it works.

src/index.ts
```diff
@@ -4,17 +4,42 @@ import cors from 'cors';
...
 
 const { ApolloServer } = require('apollo-server-express');
+
 initClients().then(({ pgClient, cloudinaryClient }) => {
     const app = express();
 
-    app.use('/status', (req, res) => {
-        res.sendStatus(200);
+    app.use('/status', async (req, res) => {
+        let isDatabaseActive = false
```
ignasposka, line 19: No semicolon in the end of the line. Please apply this for all lines where required.

src/index.ts
```diff
@@ -4,17 +4,42 @@ import cors from 'cors';
...
+        let isCloudinaryClientActive = false
+
+        try {
+            const results = await pgClient.query({
+                text: 'select true as OK'
```
ignasposka, line 24: SQL keywords have convention to be uppercase. Therefore it should be `SELECT true AS ok`.

ignasposka (approved):
Please fix code according to my comments and it will be good to merge :)

ignasposka:
Furthermore, I didn't notice code conflicts. Please fix them.

PR #191, opened 2021-04-12
#188 expand /status endpoint with more information

kostamir:


src/getCurrentVersion.ts
```diff
@@ -0,0 +1,6 @@
+import fs from 'fs';
+
+export default async function getCurrentVersion() {
+    const packageJson = await fs.promises.readFile('./package.json', 'utf8');
```
psaulenas, line 4: You could get the version using import from `package.json`
```js
```

src/index.ts
```diff
@@ -7,14 +7,32 @@ import jwks from 'jwks-rsa';
...
+        try {
+            dbStatus = (
+                await pgClient.query({
+                    text: 'SELECT VERSION()',
+                })
+            ).rows?.[0];
+        } catch (error) {
+            console.log(error);
+        }
+        const cloudinaryStatus = await cloudinaryClient.isOk();
```
JustasBelevicius, line 28: cloudinary isn't configured when the tests are run, therefore this throws an exception and the response is not 200

JustasBelevicius: try/catch should fix it

src/getCurrentVersion.ts
kostamir: Done. Thanks!

src/index.ts
kostamir: Done. Thanks!

ignasposka:

PR #192, opened 2021-04-12
#188 Expand /status endpoint with more information 

linaskitm:


src/getVersion.ts
```diff
@@ -0,0 +1,6 @@
+import fs from 'fs';
+
+export default async function getVersion() {
+    const result = await fs.promises.readFile('package.json', 'utf8');
```
psaulenas, line 4: You could get the version using import from `package.json`
```js
```

linaskitm: Thanks for review, I refactored the code, but I had to add  in tsconfig.json "resolveJsonModule": true

psaulenas: That is fine 👍

ignasposka:

PR #193, opened 2021-04-12
#188 - Expand /status endpoint with more information

lukaskairys:


src/utils/getAppVersion.ts
```diff
@@ -0,0 +1,7 @@
+import fs from 'fs';
+
+export default async function getAppVersion() {
+    const packageJson = await fs.promises.readFile('./package.json', 'utf8');
```
psaulenas, line 4: You could get the version using import from `package.json`
```js
```

ignasposka:

PR #194, opened 2021-04-13
#188-Expand status endpoint with more information.

pergaliuke:


src/index.ts
```diff
@@ -7,14 +7,27 @@ import jwks from 'jwks-rsa';
...
 
-    app.use('/status', (req, res) => {
-        res.sendStatus(200);
+    app.use('/status', async (req, res) => {
+
+        const databaseStatus = (await pgClient.query({
+            text: 'SELECT TRUE AS OK'
+        })).rows?.[0].ok;
+
+        const cloudinaryStatus = await cloudinaryClient.checkStatus();
```
psaulenas, line 23: Probably integration tests won't run because cloudinary would throw an error. The solution could be to use a try-catch block for a cloudinary check.

pergaliuke: fixed

ignasposka:

Issue #146, opened 2021-03-28
Create query for fetching organization tasks


ignasposka:
### **Goal**
Create a GraphQL query that would return all animal organization tasks from the database.

### **Description**

Two queries should be created: `organizationTasks` and `organizationTask`
Organization tasks are being stored in `organization_task` database table .

### organizationTasks

Query should be called `organizationTasks`.
`organizationTasks` query should return list of objects containing `id`, `title`, `description`, `organization` and `isDone`. Tests should be written for this functionality as well.

### Usage example
**Expected query**:
```graphql
{
  organizationTasks {
    id,
    title,
    description,
    organization,
    isDone
  }
}
```

**Expected response** :
```JSON
{
    "data": {
      "organizationTasks": [
        {
          "id": "1",
          "title": "Take away trash",
          "description": "Trash should be put in the container. Container can be found near the southern wall of the shelter",
          "organization": 2,
          "isDone": false
        },
        {
          "id": "2",
          "title": "Pet dog Rudolf",
          "description": "Dog Rudolf starts to bark everyday at 12:00 if it doesn't get pat in the morning. Barking never stops.",
          "organization": 2,
          "isDone": false
        },
        {
          "id": "3",
          "title": "Order cat food",
          "description": "Cat food is running out. Order ASAP!",
          "organization": 2,
          "isDone": true
        }
      ]
    }
  }
```

### organizationTask

Query for getting singular task also should be added. It should be `organizationTask(id: Int)`. Id here is an id of task.
`organizationTask` query should return object containing `id`, `title`, `description`, `organization` and `isDone`. Tests should be written for this functionality as well.

### Usage example
**Expected query**:
```graphql
{
  organizationTask(id: 1) {
    id,
    title,
    description,
    organization,
    isDone
  }
}
```

**Expected response** :
```JSON
{
    "data": {
      "organizationTask":
        {
          "id": "1",
          "title": "Take away trash",
          "description": "Trash should be put in the container. Container can be found near the southern wall of the shelter",
          "organization": 2,
          "isDone": false
        }
    }
  }
```

PR #200, opened 2021-04-19
#146 create query for fetching organization tasks

kostamir:


ignasposka:
Code is good, but we have already merged this feature from another PR.

PR #202, opened 2021-04-19
#146, query for fetching organization tasks

ieva1919:


ignasposka:
Code is good, but we have already merged this feature from another PR.

Issue #130, opened 2021-03-21
Create query to fetch former animal owner


ignasposka:
**Goal**

* Create a GraphQL query `formerAnimalOwners` that would return former animal owners from the database.
* Create a GraphQL query `formerAnimalOwner(id: Int!)` that would return former animal owner from the database by specific ID.

**Description**

Former animal owners are being stored in `former_animal_owner` table. 
`formerAnimalOwners` query should return list of objects each containing data from `former_animal_owner` table. Tests should be written for this functionality as well.
`formerAnimalOwner(id: Int!)` should accept former animal owner id and return single object with data from `former_animal_owner` table selected by provided ID. There are a bunch of examples one can look at. For example, we can look at organization GraphQL definition at https://github.com/pets-oss/pets-back/blob/main/src/schema/typeDefs/organization.graphql. The pull-request that introduced such a query for organizations can be seen here #79 

### Usage example:

**Expected query**:
```graphql
{
  formerAnimalOwners{
    id,
    name,
    surname,
    phone
  }
}
```

**Expected response**: (`...` means that list goes on)
```JSON
{
  "data": {
    "formerAnimalOwners": [
      {
        "id": 1,
        "name": "Rimas",
        "surname": "Karpavičius",
        "phone": "+37068745124"
      },
      {
        "id": 2,
        "name": "Algirdas",
        "surname": "Motiejūnas",
        "phone": "+37068745477"
      },
      {
        "id": 3,
        "name": "Tomas",
        "surname": "Nausėda",
        "phone":"+37068735478"
      },
      ...
    ]
  }
}
```

**Expected query**:
```graphql
{
  formerAnimalOwner(id: 1){
    id,
    name,
    surname,
    phone
  }
}
```

**Expected response**:
```JSON
{
  "data": {
    "formerAnimalOwner": {
        "id": 1,
        "name": "Rimas",
        "surname": "Karpavičius",
        "phone": "+37068745124"
    }
  }
}
```

PR #203, opened 2021-04-20
#130 Create query to fetch former animal owner

mantasimkus:
Created GraphQL queries that fetch former animal owner(s). Wrote the respective integration tests. Solves issue #130.


src/sql-queries/formerAnimalOwner.ts
```diff
@@ -0,0 +1,25 @@
+import { QueryConfig } from 'pg';
+
+export const getFormerAnimalOwnersQuery = (): QueryConfig => {
+    const text = `
+        SELECT *
```
ignasposka, line 5: This could fit into one line I think.

ignasposka (approved):
Good job! Please fix my comment and I will merge this PR.

AzisK (approved):
Good job

src/sql-queries/formerAnimalOwner.ts
mantasimkus: Done.

PR #206, opened 2021-04-20
#130 Create query to fetch former animal owner

### LINKED ISSUE #130: Create query to fetch former animal owner

LukasKri:
I've created GraphQL queries to fetch information about former animal owner.

I would be happy if I receive any comments.

Resolves #130


test/interfaces/formerAnimalOwner.interface.ts
```diff
@@ -0,0 +1,6 @@
+export default interface FormerAnimalOwner {
+    id: number;
+    name: string;
+    surname?: string;
```
ignasposka, line 4: I know that I have shown this way to defined optional fields during lecture. But to be consistent, let's use `string | null` (See other interfaces for example)

ignasposka:
Closing this PR, because another with same functionality was merged.

AzisK:
This code could have been merged as well since everything seems correct. Thanks and keep up the good work!

PR #207, opened 2021-04-20
#129 Create query for fetching available municipalities

### LINKED ISSUE #129: Create query for fetching available municipalities
**Goal**

Create a GraphQL query that would return all available municipalities from the database.

**Description**

Municipality ID is required when creating animal found event (https://github.com/pets-oss/pets-back/issues/112).
This ticket should introduce new query to fetch available municipalities.

Municipalities are being stored in `municipality` table. `municipalities` query should return list of objects each containing `id` and `value`. Tests should be written for this functionality as well.

### Usage example

**Expected query**:
```graphql
{
  municipalities {
    id,
    value
  }
}
```

**Expected response** (`...` means that list goes on):
```JSON
{
    "data": {
      "municipalities": [
        {
          "id": "1",
          "value": "Akmenės rajono savivaldybė"
        },
        {
          "id": "2",
          "value": "Alytaus miesto savivaldybė"
        },
        {
          "id": "3",
          "value": "Alytaus rajono savivaldybė"
        },
        {
          "id": "4",
          "value": "Anykščių rajono savivaldybė"
        },
        ...
      ]
    }
  }
```

LukasKri:
Created GraphQL query that returns all available municipalities.

Resolves #129


ignasposka (approved):
Good job!

PR #209, opened 2021-04-22
#119 Add query to get all available chip install places

### LINKED ISSUE #119: Add query to get all available chip install places
**Goal**

Create a  GraphQL query that would return all available chip install places from the database.

**Description**

Front-end needs to be able to retrieve all possible chip install places because it needs to show them for a user when entering animal microchip installation information. Please create such a GraphQL query and write automatic tests that would ensure it works correctly. There are a bunch of examples one can look at. For example, we can look at gender GraphQL definition at https://github.com/pets-oss/pets-back/blob/main/src/schema/typeDefs/gender.graphql. The pull-request that introduced such a query for genders can be seen here #47

Thus our chip install places GraphQL definition could look like the following

```graphql
extend type Query {
    """
    Get all chipInstallPlaces.

    Examples:

    chipInstallPlaces(language: "lt")
    """
    chipInstallPlaces("language code" language: String!): [ChipInstallPlace]
}

"Represents a chip install place."
type ChipInstallPlace {
    "Chip install place id"
    id: String!
    "Chip install place value"
    value: String!
}
```

thus

**Expected behaviour**

Such GraphQL query

```graphql
{
    chipInstallPlaces(language: "lt") {
        id,
        value
    }
}
```

Should return all chip install places with ids and values in Lithuanian language

LukasKri:
Created GraphQL query to get all available chip install places.

Resolves #119


src/schema/typeDefs/chipInstallPlace.graphql
```diff
@@ -0,0 +1,18 @@
...
+    Examples:
+
+    chipInstallPlaces(language: "lt")
+    """
+    chipInstallPlaces("language code" language: String!): [ChipInstallPlace]
+}
+
+"Represents a chip install place."
+type ChipInstallPlace {
+    "Chip install place id"
```
psaulenas, line 14: It is not place id but translation id

src/sql-queries/chipInstallPlace.ts
```diff
@@ -0,0 +1,18 @@
+import { QueryConfig } from 'pg';
+
+const getChipInstallPlacesQuery = (language: string): QueryConfig => {
```
psaulenas, line 3: I would name query something like `getChipInstallPlacesTranslationQuery`

src/schema/typeDefs/chipInstallPlace.graphql
LukasKri: Thanks, you're right.

src/sql-queries/chipInstallPlace.ts
```diff
@@ -0,0 +1,20 @@
+import { QueryConfig } from 'pg';
+
+const getChipInstallPlacesTranslationQuery = (
+    language: string
+): QueryConfig => {
+    const text = `
+        SELECT
+            install_place AS id,
+            translation AS value
+        FROM install_place_translation
```
psaulenas, line 10: If you don't mind, could you rename this table a little bit?
I think it is worth adding `chip` in the beginning.

src/sql-queries/chipInstallPlace.ts
```diff
@@ -0,0 +1,20 @@
+import { QueryConfig } from 'pg';
```
psaulenas, line 1: Could you rename the file to match the table name that you are using?

src/sql-queries/chipInstallPlace.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const getChipInstallPlacesTranslationQuery = (
+    language: string
+): QueryConfig => {
+    const text = `
+        SELECT
+            install_place AS id,
```
psaulenas, line 8: And mind it would be more clear if we would rename this column to something like `chip_install_place_id`. What do you think?

test/chipInstallPlace.graphql.test.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('GraphQL chip install place integration tests', () => {
+    it('Returns all chip install places in "lt" with all fields', (done) => {
+        request
+            .post('/graphql')
+            .send({
+                query: `{ chipInstallPlaces(language: "lt")
+                    ${translationFields}
+                }`,
+            })
+            .set('Authorization', `Bearer ${process.env.BEARER_TOKEN}`)
```
psaulenas, line 20: Nitpick, but I think it is not needed to pass the authorization header in integration tests because authentication was turned off in tests.

src/schema/typeDefs/chipInstallPlace.graphql
```diff
@@ -0,0 +1,18 @@
...
+    Examples:
+
+    chipInstallPlaces(language: "lt")
+    """
+    chipInstallPlaces("language code" language: String!): [ChipInstallPlace]
+}
+
+"Represents a chip install place."
+type ChipInstallPlace {
+    "Chip install place translation id"
```
psaulenas, line 14: Sorry, I was wrong. Actually, this `id` is not a translation but `install_place`. [Check the schema](https://github.com/pets-oss/pets-back/blob/1c503d26d9b86bdbb11833e01bdc252165454a73/database/1-schema.sql#L181-L188) that it would be more clear.

psaulenas (commented):
Good job!
Sorry, I was wrong regarding the translation id yesterday. I think it would be good to improve naming a little bit that it would be less confusing.

src/sql-queries/chipInstallPlace.ts
LukasKri: No worries! Just to be clear, you want me to make these changes in the `1-schema.sql` file first and then update it accordingly in this SQL query, am I right?

src/sql-queries/chipInstallPlace.ts
LukasKri: Of course, I'm sorry for not following the convention. The right name for this file would be `installPlaceTranslation.ts`, right? I suppose I should I update the name of every file regarding this PR in accordance with this rule to maintain consistency.

src/sql-queries/chipInstallPlace.ts
LukasKri: I agree with your opinion on this one, but I'm still a little confused about your first comment about adding `chip` in the beginning. Could you please elaborate more where do you want me to add the word `chip`?

test/chipInstallPlace.graphql.test.ts
LukasKri: Well if that line of code is useless I'll remove it, that's not a problem at all. So if the authentication process is turned off in testing, then `.set()` method, that is used in every test, is kind of redundant and do not bring any value, right?

And thank you for nitpicking, it helps me to understand the code better and testing in general.

src/schema/typeDefs/chipInstallPlace.graphql
LukasKri: Well according to the schema `install_place` is ENUM, which have the values from 1 to 4 and these values translate to a specific chip install place. In our case, the `id` represents `install_place` ENUM value and there are no `translation_id` definition in this table, so that's why we should actually call it "Chip install place id", correct?

src/sql-queries/chipInstallPlace.ts
psaulenas: I had in mind to rename the table name. So, in order to do that, you would need to make changes in the `1-schema.sql`.
The reason why I suggest renaming is that is not clear that this table stores information about the chip. So, I think `chip_ install_place_translation` name would be more appropriate.

src/schema/typeDefs/chipInstallPlace.graphql
psaulenas: Yes, that is correct

src/sql-queries/chipInstallPlace.ts
psaulenas: Yeah, I would expect something similar to that. If you rename the table, then it would be `chipInstallPlaceTranslation.ts.`

src/sql-queries/chipInstallPlace.ts
psaulenas: If we rename the table to `chip_install_place_translation`, then we don't need `chip` in the column name. But I think it would be great to add  ending `_id`, so then the column name could be `install_place_id`

test/chipInstallPlace.graphql.test.ts
psaulenas: OK, I will add the context why this header is set in other tests. Initially, we had tests that were using authorization and we used Github secrets store for that. However, when the PR is created from the fork, then the secrets are not passed to the Github action and tests fail. So, the authorization was turned off for integration tests.

src/sql-queries/chipInstallPlace.ts
LukasKri: Yes, that makes sense, I've changed the name of the table to `chip_install_place_translation` and updated other related parts of the code to match it. I didn't think we are allowed to make changes to the database schema so simply though, but that's because the project is not deployed yet, right?

The name of the table was updated in these files: `1-schema.sql`, `2-data.sql` and `chipInstallPlaceTranslation.ts`.

src/sql-queries/chipInstallPlace.ts
LukasKri: I've renamed all files according to the new name of the table as you suggested (`chipInstallPlaceTranslation`).

src/schema/typeDefs/chipInstallPlace.graphql
LukasKri: Thank you, @psaulenas, for your time and feedback, I really appreciate that.

src/sql-queries/chipInstallPlace.ts
psaulenas: The database is cleaned before every deployment to dev and prod environments currently.
You are right, if this application was used, we couldn't modify just the schema file because we couldn't delete all the data. In that case, a separate migration file should be created to apply the schema changes.

test/chipInstallPlace.graphql.test.ts
psaulenas: Actually, even collaborators don't have access to the secret that is set. They can just update the secret value but can't see the value. The secrets are not passed to PRs that are created from fork to prevent leaking them, for example, https://stackoverflow.com/a/63009069

psaulenas (approved):
LGTM, but I think there are some unintentional changes

LukasKri (commented):
@psaulenas actually changes in `3-fake-data.sql` and `reports.sql` were unintentional and I found one more in `organizationTask.graphql.test.ts` file that I didn't make, so I reverted all of them. I do not know how to explain these changes and I was unaware about them until you pointed them out to me. I didn't recall having any conflicts, my apologies for not paying attention. I think I managed to revert these changes back, could you have a look at it?

psaulenas (approved):
@LukasKri GJ!
However, there is a conflict. Could you please resolve it? Then I will merge a PR

LukasKri:
Thanks for your help and feedback @psaulenas, I've resolved the conflict for the first time and it feels great! Have a good day.

psaulenas (approved):
Good job!

Issue #132, opened 2021-03-21
Create Date scalar


ignasposka:
Date scalar should be string matching `YYYY-MM-DD` format. It should throw exception in case of format is violated.

You can follow example defined in https://www.apollographql.com/docs/apollo-server/schema/custom-scalars/

pergaliuke:
Assign me, please

PR #210, opened 2021-04-23
 #132 Create Date scalar

pergaliuke:
Date scalar is string matching YYYY-MM-DD format, throws Error in case of format violation or invalid date (e.g. 2021-02-30).


ignasposka (changes requested):
Well done solution! Can we have some tests for `src/schema/scalars/Date.ts`?

pergaliuke:
> Well done solution! Can we have some tests for `src/schema/scalars/Date.ts`?

Thank you. Added some tests.

ignasposka (approved):
Well done solution, thanks!

PR #212, opened 2021-04-27
Feature/#131 add mutations for former animal owner

pergaliuke:
Added scalars Name, Surname, Phone.
Added mutations for former animal owner.
Added tests for mutations.
Minor Fix in former_animal_owner table.

I am not sure if I placed scalars in a correct folder or if my DRY approach was good.

I am little bit confused with required fields, does it mean not null? I might have made a mistake there.


src/schema/scalars/getStringScalarWithValidation.ts
```diff
@@ -0,0 +1,43 @@
+import { GraphQLScalarType, Kind } from 'graphql';
+
+interface Config {
+    name: string,
+    max?: number,
```
ignasposka, line 5: Let's use `| null` to be consistent with other interfaces in Petbook project.

src/schema/scalars/scalars.ts
```diff
@@ -0,0 +1,9 @@
+import getStringScalar from './getStringScalarWithValidation';
+
+const nameScalar = getStringScalar({ name: 'Name', max: 255 });
+const surnameScalar = getStringScalar({ name: 'Surname', max: 255 });
+const phoneScalar = getStringScalar({
+    name: 'Phone', pattern: /^\+3706\d{7}$/
```
ignasposka, line 6: I think we should move country code to a config file, by doing that we would let use system not only with Lithuanian phone numbers.

src/schema/typeDefs/formerAnimalOwner.graphql
```diff
@@ -28,3 +28,28 @@ type FormerAnimalOwner {
...
+    surname: Surname!
+    "Former animal owner phone number (+3706XXXXXXX)"
+    phone: Phone
+}
+
+input UpdateFormerAnimalOwnerInput {
+    "Former animal owner ID, e.g., 1"
+    id: Int!
+    "Former animal owner name (255 characters max)"
+    name: Name!
```
ignasposka, line 50: Name shouldn't be required during update, maybe user would want to change some other fields and leave name untouched.

src/sql-queries/formerAnimalOwner.ts
```diff
@@ -1,18 +1,49 @@
 import { QueryConfig } from 'pg';
+import { insert, update } from 'sql-bricks-postgres';
+import snakeCaseKeys from 'snakecase-keys';
+
+const table = 'former_animal_owner';
+const returnFields = '*';
```
ignasposka, line 6: I think moving `*` to a variable is pointless here, because it's well known SQL character, furthermore it would be more readable if it would be in a query.

test/formerAnimalOwner.graphql.test.ts
```diff
@@ -41,4 +46,54 @@ describe('GraphQL former animal owner integration tests', () => {
...
+                    }`,
+            })
+            .expect(200)
+            .set('Authorization', `Bearer ${process.env.BEARER_TOKEN}`)
+            .end((err, res) => {
+                if (err) return done(err);
+                const { body: { data: { updateFormerAnimalOwner } } } = res;
+                validate(updateFormerAnimalOwner);
+                expect(updateFormerAnimalOwner).to.include({
+                    id: 2,
```
ignasposka, line 93: I would suggest to save id from create mutation response and use it here to update freshly created entry. By that we always would be sure that id exist.

ignasposka (changes requested):

ignasposka:
> I am little bit confused with required fields, does it mean not null? I might have made a mistake there.

Yes, required field means it won't be null.

src/schema/scalars/getStringScalarWithValidation.ts
pergaliuke:
```
src/schema/scalars/scalars.ts(5,37): error TS2345: Argument of type '{ name: string; pattern: RegExp; }' is not assignable to parameter of type 'Config'.
  Property 'max' is missing in type '{ name: string; pattern: RegExp; }' but required in type 'Config'.
```

`max: number | null ` will not compile with optional fields. If we change optional(?) to | null or | undefined,  field becomes **required** with possible values `null `or `undefined`.

`max?: number` is more like `max: number | undefined` in **strict** TypeScript, but not exactly due to required or not.

More information:
https://www.typescriptlang.org/docs/handbook/2/objects.html#property-modifiers.

ignasposka: That's a good point, thanks for clarifying.

test/formerAnimalOwner.graphql.test.ts
pergaliuke: Done

src/schema/typeDefs/formerAnimalOwner.graphql
pergaliuke: Fixed.

src/schema/scalars/scalars.ts
pergaliuke: Changed, please look at this again. Pattern was : /^\+3706\d{7}$/ , now pattern allows more digits than Lithuania has after country code. It is ok?

src/schema/scalars/getStringScalarWithValidation.ts
```diff
@@ -36,7 +36,7 @@ const getStringScalarWithValidation = (config: Config) =>
             if (ast.kind === Kind.STRING) {
                 return validate(ast.value, config);
             }
-            return null;
+            throw new Error('Value must be a string');
```
ignasposka, line 39: Let's throw `ValidationError` instead of `Error`.

src/schema/scalars/getStringScalarWithValidation.ts
```diff
@@ -0,0 +1,43 @@
+import { GraphQLScalarType, Kind } from 'graphql';
+
+interface Config {
+    name: string,
+    max?: number,
+    pattern?: RegExp
+}
+
+const validate = (value: unknown, config: Config) => {
```
ignasposka, line 9: I think we should add test for this validation.

src/schema/scalars/scalars.ts
```diff
@@ -0,0 +1,11 @@
+import getStringScalar from './getStringScalarWithValidation';
+import config from '../../config';
+
+const nameScalar = getStringScalar({ name: 'Name', max: 255 });
+const surnameScalar = getStringScalar({ name: 'Surname', max: 255 });
+const phoneScalar = getStringScalar({
+    name: 'Phone',
+    pattern: new RegExp(`^\\+${config.phoneCountryCode}\\d{8,10}$`)
```
ignasposka, line 8: Add default country code fallback in case one from config is missing.

src/config.ts
```diff
@@ -0,0 +1,3 @@
+export default {
```
ignasposka, line 1: I would have put this into some kind of `.env` file. But I'm not sure if that's a right way or not. @JustasBelevicius @AzisK @JBK91 @psaulenas  what do you think?

ignasposka (changes requested):

src/config.ts
psaulenas: I don't think this should be stored in the environment variables. I believe we should think about an easily extensible solution. Also, we should support other countries phone formats but probably default to `LT`. So, my suggestion is to have a configuration something like:

```js
export phoneFormatByCountry {
 lt: '370'
 lv: '371'
}
```

Probably, you will get a lint error because there is only one export in this file at the moment. However, I would disable that lint error. And personally, I don't like that lint rule.

ignasposka: Alright, I like Paulius suggestion about the format.
But for default exports let's keep them consistent - let's have default one if it's a single export.

ignasposka (changes requested):

ignasposka (approved):
Looks good. You were asking about testing case while passing non string value into string scalar, I think you can add this test case expecting it to throw `ValidationError`, I don't see a problem here.

pergaliuke:
> Looks good. You were asking about testing case while passing non string value into string scalar, I think you can add this test case expecting it to throw `ValidationError`, I don't see a problem here.

I wanted to do something like 
const INVALID = [ 0, 3.2, true] 
I am little bit confused about about null values. It seems that null does not go through Scalar at all.

AzisK (approved):
Very good job!

PR #214, opened 2021-05-02
Add input validation for animal mutations

psaulenas:


src/schema/resolvers/animal.ts
```diff
@@ -79,6 +80,28 @@ const resolvers: IResolvers = {
     },
     Mutation: {
         createAnimal: async (_, { input }, { pgClient, cloudinaryClient }) => {
+            const createAnimalInputValidator = new Validator(input, {
+                name: 'maxLength:128',
+                organization: 'integer|min:1',
```
ignasposka, line 85: Organization should be required.

src/schema/resolvers/animal.ts
```diff
@@ -133,6 +158,28 @@ const resolvers: IResolvers = {
             }
         },
         updateAnimal: async (_, { input }, { pgClient, cloudinaryClient }) => {
+            const updateAnimalInputValidator = new Validator(input, {
```
ignasposka, line 161: Id could be validated as well.

ignasposka (approved):

src/schema/resolvers/animal.ts
psaulenas: I don't want to double the check because this validation is done on GQL side (mandatory field)

src/schema/resolvers/animal.ts
psaulenas: It is done on GQL side

Issue #113, opened 2021-03-02
Add mutation for given away event


ignasposka:
**Goal**

* Create a GraphQL mutation `createGivenAwayEvent` that would create found event data in the database.
* Create a GraphQL mutation `updateGivenAwayEvent` that would update found event data in the database by specific ID.

**Description**

* Given away events are being stored in `animal_event_given_away` table. 
* Tests should be written for this functionality as well.

`createGivenAwayEvent` should accept fields:
| Field name  | Data Type |
| ------------- | ------------- |
| formerOwnerId  | Int (Required) |
| reason | String |
| animalId | Int (Required) |
| date | Date |

`updateGivenAwayEvent` should accept fields:
| Field name  | Data Type |
| ------------- | ------------- |
| id | Int (Required) |
| formerOwnerId  | Int |
| reason | String |
| animalId | Int |
| date | Date |

For example, we can look at organization GraphQL definition at https://github.com/pets-oss/pets-back/blob/main/src/schema/typeDefs/organization.graphql. The pull-request that introduced such mutations for organizations can be seen here #79 

### Usage example:

**Expected mutation**:
```graphql
mutation {
  createGivenAwayEvent (input: {
    formerOwnerId: 5,
    reason: "Leaving country"
    date: "2021-03-19",
    animalId: 4,
  }) {
    id,
    formerOwnerId,
    date,
    animalId,
    reason
  }
}
```
**Expected response** (returned id does not need to be exactly 1):
```JSON
{
  "data": {
    "createFoundEvent": {
      "id": 1,
      "formerOwnerId": 5,
      "reason": "Leaving country",
      "date": "2021-03-19",
      "animalId": 4,
    }
  }
}

```

**Expected mutation**:
```graphql
mutation {
  updateGivenAwayEvent (input: {
    id: 1,
    formerOwnerId: 6,
    reason: "Allergy"
    date: "2021-03-20",
    animalId: 5,
  }) {
    id,
    formerOwnerId,
    date,
    animalId,
    reason
  }
}
```
**Expected response**:
```JSON
{
  "data": {
    "createFoundEvent": {
      "id": 1,
      "formerOwnerId": 6,
      "reason": "Allergy",
      "date": "2021-03-20",
      "animalId": 5,
    }
  }
}

```

pergaliuke:
I would like to work on this task. Assign me, please

PR #220, opened 2021-05-03
#113 Add mutation for given away event

pergaliuke:
Added mutations createGivenAwayEvent ant updateGivenAwayEvent  with tests.

1. **date** field is yyyy-mm-dd, but in database **date_time** timestamp, should I change something?
2. Should I add check in resolver for updateGivenAwayEvent  if fields formerOwnerId and animalId **are not null**, any other additional checks needed?
3. Should I add **queries** or **delete mutation** as well?


ignasposka:
> Added mutations createGivenAwayEvent ant updateGivenAwayEvent with tests.
> 
> 1. **date** field is yyyy-mm-dd, but in database **date_time** timestamp, should I change something?
> 2. Should I add check in resolver for updateGivenAwayEvent  if fields formerOwnerId and animalId **are not null**, any other additional checks needed?
> 3. Should I add **queries** or **delete mutation** as well?

1. Let's change database field to be `DATE`. Given away time is not that important, main advantage is to know which day animal was given away.
2.  No, on update no fields should be required except from `id`. You can only check if input object contains more than one field, because there is no point of performing update operation if user providing only `id`, but no fields to update.
3. Query will be added later, as we trying to develop general query to return all types of events. If you wish you can add delete mutation as well, but we are fine to doing that in a separate issue.

ignasposka (changes requested):
Requesting changes to change database field from `TIMESTAMP` to `DATE`

pergaliuke:
> Requesting changes to change database field from `TIMESTAMP` to `DATE`

Done.

ignasposka (approved):

PR #223, opened 2021-05-03
Feature/#112 create mutation for animal found event

pergaliuke:
Added mutation createFoundEvent with validation and tests

### DIFF
--- a/src/schema/resolvers/animalEventFoundResolver.ts
+++ b/src/schema/resolvers/animalEventFoundResolver.ts
@@ -1,5 +1,6 @@
 import { IResolvers } from 'graphql-tools';
 import { Validator } from 'node-input-validator';
+import { ValidationError } from 'apollo-server-express';
 import {
     getAnimalFoundEventsQuery,
     createAnimalFoundEventQuery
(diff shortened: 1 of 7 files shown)


src/schema/resolvers/animalEventFoundResolver.ts
```diff
@@ -8,8 +12,32 @@ const resolvers: IResolvers = {
...
+                houseNo: 'maxLength:8',
+                municipalityId: 'integer|min:1',
+                animalId: 'integer|min:1',
+                date: 'date|dateBeforeToday:0,days'
+            });
+            const isCreateFoundEventInputValid =
+                await createFoundEventInputValidator.check();
+
+            if (!isCreateFoundEventInputValid) {
+                throw new Error(
```
ignasposka, line 30: Let's use `ValidationError` here, it would be more appropriate.

ignasposka (changes requested):
Sorry, for not reviewing your PR so long. Please change that type of error I have commented about and I will merge PR.

src/schema/resolvers/animalEventFoundResolver.ts
pergaliuke: Done.

ignasposka (approved):

PR #238, opened 2021-05-19
events fetching

JBK91:
Will add tests later in same PR


src/schema/resolvers/event.ts
```diff
@@ -1,75 +1,86 @@
...
+const resolvers: IResolvers = {
+    Query: {
+        events: async (_, { animalId, groups }, { pgClient }) => {
+            const events = [];
 
-            return dbResponse.rows;
-        },
-        general: async (_, __, { pgClient }) => {
-            const dbResponse = await pgClient.query(getAllGeneralEvents());
+            if (!groups?.length || groups.includes('General')) {
```
ignasposka, line 62: You can use just `!groups` instead of `!groups?.length`

ignasposka (approved):

PR #239, opened 2021-05-23
Improve events schema

ignasposka:


src/schema/typeDefs/event.graphql
```diff
@@ -16,12 +16,12 @@ enum EventGroup {
...
 enum EventType {
     GivenAway,
-    Rescued,
+    Found,
     CheckIn,
     CheckOut,
     Died,
     TemporaryCare,
-    Microchipped,
+    Microchipping,
```
AzisK, line 24: I know this comes not from this PR but why do we have some words in verb past form (Died, Found), some as nouns (TemporaryCare, Microchipping) etc.? I believe they should have the same form

src/schema/resolvers/event.ts
```diff
@@ -67,19 +117,18 @@ const resolvers: IResolvers = {
                 events.push(...appendEventsDetails(generalEvents.rows))
                 events.push(...appendFoundEventsDetails(foundEvents.rows))
                 events.push(...appendGivenAwayEventsDetails(givenAwayEvents.rows))
+                events.push(...getMicrochippingEvents());
```
AzisK, line 120: This is temporary as far as I understand?

database/1-schema.sql
```diff
@@ -272,7 +272,7 @@ CREATE TYPE event_group AS ENUM('General', 'Medical');
 
 CREATE TYPE event_type AS ENUM (
     'GivenAway',
-    'Rescued',
+    'Found',
```
AzisK, line 275: I know this comes not from this PR but the events do not represent all of https://www.vic.lt/gpsas-apskaita/gar-klasifikatoriai/ event types as in https://www.vic.lt/gpsas-apskaita/wp-content/uploads/sites/6/2020/03/Ivykiu_KL_20200303.pdf

AzisK (approved):
Looks good, left some comments

src/schema/resolvers/event.ts
ignasposka: Yes, that's just a mocked data until we have required DB structure for this.

database/1-schema.sql
ignasposka: As this was done according to inside specification let's not block this PR because of that.

@saulyz can you comment why not all event types are represented in our specification? Why do we have some discrepancies?

src/schema/typeDefs/event.graphql
ignasposka: I agree that this enum lacks of consistency.
@saulyz Can we update our specification to have same form for every event?

Issue #72, opened 2021-01-22
Implement animals pagination with GraphQL


AzisK:
**Goal**

Implement animals pagination in GraphQL queries

**Description**

Implement animals pagination in GraphQL queries by putting single entries to nodes and assigning cursors to those nodes. Pagination should be able to accept such arguments:

| Field name  | Data Type |
| ------------- | ------------- |
| after  | String  |
| first | Int |
| before | String |
| last | Int |

Node and cursor should be put into an edge. There should also be pageInfo next to the list of edges. pageInfo should have such fields:

| Field name  | Data Type |
| ------------- | ------------- |
| hasNextPage  | Boolean (Required)  |
| hasPreviousPage | Boolean (Required) |
| startCursor | String |
| endCursor | String |

[...]

PR #252, opened 2021-05-26
#72 implement animals pagination with graphql. IMPORTANT. Do not add this directly to production until..

pergaliuke:


src/sql-queries/animal.ts
```diff
@@ -60,39 +60,32 @@ export const getAnimalsQuery = (
...
+        ${table}.comments,
+        ${table}.mod_time`)
+        .from(table)
+        .leftJoin('animal_details as ad').on(`${table}.id`,'ad.animal_id')
+        .leftJoin('breed AS b').on('ad.breed_id','b.id');
+    query = ids ? query.where($in(`${table}.id`, ids)) : query;
+    query = species ? query.where($in('b.species', species)) : query;
+    query = gender ? query.where($in('ad.gender_id', gender)) : query;
+    query = breed ? query.where($in('ad.breed_id', breed)) : query;
+    query = cursor ? query.where(reverse ? lt('id', cursor) : gt('id', cursor)) : query;
```
AzisK, line 84: As there are a bunch of tables that have id, I believe this part should be non-ambiguous

AzisK: Due to that, such query failed for me. You could add a test for this

```
{
  animals(first: 2, after: "MQ==") {
    edges {
      node {
        id
        name
      }
      cursor
    }
    pageInfo{
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
  }
}
```

src/sql-queries/animal.ts
```diff
@@ -60,39 +60,32 @@ export const getAnimalsQuery = (
...
+    query = reverse ? query.orderBy('id DESC') : query;
```
AzisK, line 85: I don't really want to but it seems to me that we should usually order by id ascending to be consistent and correct all the time as well

AzisK (changes requested):
Great job! Minor improvement suggested

AzisK (approved):
Thumbs up!

test/interfaces/animalConnection.interface.ts
```diff
@@ -0,0 +1,8 @@
+import Animal from './animal.interface';
+import Connection from './connection.interface';
+
+export default interface AnimalConnection extends Connection<Animal> {
```
JustasBelevicius, line 4: I believe `export default type AnimalConnection = Connection<Animal>;` would be sufficient here.

src/schema/resolvers/animal.ts
```diff
@@ -67,13 +67,64 @@ async function getUpdateMicrochipResult(input: any, pgClient: any) {
...
+                species,
+                gender,
+                breed,
+                limit ? limit + 1 : limit,
+                reverse,
+                cursor ? Buffer.from(cursor, 'base64').toString() : null
+            ));
+
+            let { rows } = dbResponse;
+            const hasMore = limit ? rows.length > limit : false;
```
JustasBelevicius, line 102: nit: I would prefer `const hasMore = limit && rows.length > limit;` but that's just my preference

test/interfaces/connection.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+    pageInfo: PageInfo
+}
+
+interface Edge<T> {
+    node: T
+    cursor: string
+}
+
+interface PageInfo {
+    startCursor : string
```
JustasBelevicius, line 12: nit: spaces before `:` are not required.

JustasBelevicius (approved):
This is very impressive! Well done 👏 ! Wait for more reviews before merging though, there is quite a lot of logic to review.

pergaliuke:
Now it returns correct results with first: 0 or last: 0 as well.

AzisK:
> Now it returns correct results with first: 0 or last: 0 as well.

What does it return in this case?

pergaliuke:
> > Now it returns correct results with first: 0 or last: 0 as well.
> 
> What does it return in this case?

```
{
  animals(first: 0) {
    ...
  }
}
```
```
{
  "data": {
    "animals": {
      "edges": [],
      "pageInfo": {
        "hasNextPage": true,
        "hasPreviousPage": false,
        "startCursor": null,
        "endCursor": null
      }
    }
  }
}
```

```
{
  animals(last: 0) {
    ...
  }
}
```
```
{
  "data": {
    "animals": {
      "edges": [],
      "pageInfo": {
        "hasNextPage": false,
        "hasPreviousPage": true,
        "startCursor": null,
        "endCursor": null
      }
    }
  }
}
```

test/interfaces/connection.interface.ts
```diff
@@ -0,0 +1,16 @@
+export default interface Connection<T> {
+    edges: Edge<T>[]
```
ignasposka, line 2: Missing semicolons. @JustasBelevicius Can we add this to eslint rules?

src/schema/typeDefs/animal.graphql
```diff
@@ -25,8 +26,44 @@ extend type Query {
         "Animal gender ids in database"
         gender: [Int]
         "Animal breed ids in database"
-        breed: [Int]
-    ): [Animal]
+        breed: [Int],
+        "Cursor to return animals after that cursor"
+        after: String,
```
ignasposka, line 31: I think it would be more clear if parameters would go in the following order: `after`, `before`, `first`, `last`.

ignasposka (changes requested):
Good job! Just added some small suggestions.

src/schema/typeDefs/animal.graphql
AzisK: I wouldn't necessarily agree because looking at SWAPI it's this way https://graphql.org/swapi-graphql/. However, looking at Githubs API it's as you suggest https://docs.github.com/en/graphql/overview/explorer. To me it makes sense that the variables that should go together are next to each other

src/schema/resolvers/animal.ts
```diff
@@ -67,13 +67,61 @@ async function getUpdateMicrochipResult(input: any, pgClient: any) {
...
+            const reverse = (last ?? before) != null
+            const limit = first ?? last;
+            const cursor = after ?? before;
+
+            const dbResponse = await pgClient.query(getAnimalsQuery(
+                ids,
+                species,
+                gender,
+                breed,
+                limit != null ? limit + 1 : limit,
```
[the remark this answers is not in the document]
pergaliuke: disagree, the result would be different then limit is 0

src/schema/typeDefs/animal.graphql
pergaliuke: I haven't changed this one yet. AzisK point looks very convincing.

ignasposka: Alright, I don't have a strong preference about this. If you say it's not more clear, let's leave it as it is.

src/schema/resolvers/animal.ts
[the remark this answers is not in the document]
ignasposka: Good catch!

ignasposka (approved):
Impressive work 👏 Thank you!

ignasposka:
@pergaliuke Looks like there is a TS error in `favoriteAnimal.ts` file after updating branch. Can you take a look?

src/schema/resolvers/animal.ts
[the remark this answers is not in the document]
AzisK: Do you think it's worth writing a test for such case?

Issue #240, opened 2021-05-24
Error output is not being shown in docker logs


ignasposka:
How to reproduce:
1. Go to any `.graphql` file and add duplicate field for any type
2. Go to docker log and see `[nodemon] app crashed - waiting for file changes before starting...`

Expected output in docker log is: `Error: Field "FoundEvent.author" can only be defined once.`

pergaliuke:
> How to reproduce:
> 
> 1. Go to any `.graphql` file and add duplicate field for any type
> 2. Go to docker log and see `[nodemon] app crashed - waiting for file changes before starting...`
> 
> Expected output in docker log is: `Error: Field "FoundEvent.author" can only be defined once.`

Is this is still an issue? I can not reproduce:
![image](https://user-images.githubusercontent.com/48152843/120179207-f32dd880-c212-11eb-990c-8bec7700af27.png)

By default, Apollo Server omits the exception.stacktrace field if the NODE_ENV environment variable is set to either production or test.

AzisK:
I also get this error although I don't have NODE_ENV anywhere

AzisK:
Closing unless someone can confirm this issue

PR #254, opened 2021-05-28
#228_Create_resolver_for_app_user

pergaliuke:
Not finished, wanted to check if solution would be good enough. As well I hoped to use inheritResolversFromInterfaces, but it doesn't work.


pergaliuke:
Error output is not being shown in docker logs #240.
addResolversToSchema handles schema validation, errors are visible in docker logs. For example.
image

src/schema/index.ts
```diff
@@ -52,8 +56,9 @@ const schema = loadSchemaSync('src/schema/typeDefs/*.graphql', {
         animalFoundEventResolver,
         givenAwayEventResolvers,
         favoriteAnimalResolvers,
+        givenAwayEventResolvers,
```
AzisK, line 59: I see `givenAwayEventResolvers` 2 times

src/schema/typeDefs/animalRegistration.graphql
```diff
@@ -3,6 +3,10 @@ enum RegistrationStatus {
     Inactive
 }
 
+extend type Query {
```
AzisK, line 6: Is this part of this issue?

test/event.graphql.test.ts
```diff
@@ -7,6 +7,16 @@ require('dotenv').config({ path: './test/.env' });
 const url = process.env.TEST_URL || 'http://localhost:8081';
 const request = supertest(url);
 
+// eslint-disable-next-line import/prefer-default-export
+export const authorFields = `
```
AzisK, line 11: Maybe this is a good candidate to go into a separate file?

AzisK (changes requested):
Maybe we should not return all the information but otherwise it looks good to me.

Left a few comments and questions

src/schema/typeDefs/animalRegistration.graphql
pergaliuke: After I changed to addResolversToSchema, app crashed
image

I fixed it by adding to schema. But maybe we can remove it from resolvers and from schema if this functionality is not needed anymore.

src/schema/index.ts
pergaliuke: Good catch :)

test/event.graphql.test.ts
pergaliuke: Done

src/index.ts
```diff
@@ -82,8 +82,23 @@ initClients().then(({ pgClient, cloudinaryClient }) => {
...
         schema,
+        formatResponse: (response: any) => {
+            if (response.errors) {
+                console.log(response);
+                // eslint-disable-next-line default-case
+                switch (response.errors[0].extensions.code) {
+                case 'GRAPHQL_VALIDATION_FAILED':
+                case 'BAD_USER_INPUT':
+                    console.log(response?.http?.status);
+                    // response.http.status.set(400);
```
JustasBelevicius, line 93: Well done! One note here: either remove the console statement, or remove the commented out line.

JustasBelevicius (approved):

AzisK (approved):

Issue #245, opened 2021-05-24
Create table for Medication event details


AzisK:
**Goal**
Create a table for Medication event details

**Description**
Suggested table structure 

| Column name | Type |
| - | -|
| id | Serial |
| event_id | Int (FK to events table id column) |
| treatment | Varchar |
| expenses | Number |

like in this mock https://github.com/pets-oss/pets-back/blob/da4a7f791f355c7e3e2d84703032c4cfcec6bf57/src/schema/resolvers/event.ts#L68-L83 `comments`will be something that every event in `events` table will have. Please also add some fake data.

This issue depends on this issue https://github.com/pets-oss/pets-back/issues/250. Table `events` to store all events is first needed to then link (by foreign key) event details table to it.

Similar issue https://github.com/pets-oss/pets-back/issues/244

PR #256, opened 2021-06-02
#245 Create table for medication event details

linaskitm:


AzisK (approved):
GJ

database/3-fake-data.sql
```diff
@@ -162,3 +162,9 @@ INSERT INTO animal_cage (animal_id, cage_id) VALUES
 (2, 4),
 (3, 5),
 (4, 6);
+
+INSERT INTO medication_event_details (event_id, treatment, expenses) VALUES
+(1, 'Some pills from Prophylaxis', 20.00),
```
AzisK, line 167: Medication event in `events` table is only the third one. Others are not really medication events. You would need to add them in `events` table.
`(3, 'Medical', 'Medication', '2020-05-21', 'dhjbwau74a6', 'Tick medications')`

AzisK (changes requested):
Medication event in `events` table is only the third one. Others are not really medication events. You would need to add them in `events` table.
`(3, 'Medical', 'Medication', '2020-05-21', 'dhjbwau74a6', 'Tick medications')`

database/3-fake-data.sql
linaskitm: Sorry for late response.
Now I see why we needed these changes, we will create tables for all `*_event_details`

AzisK (approved):
GJ

Issue #257, opened 2021-06-02
Enhance Connection


pergaliuke:
Make Node not null
Make Edges not null
Add Total to PageInfo

PR #258, opened 2021-06-02
#257 Enhance AnimalConnection

pergaliuke:
totalCount returns **all animals count**, and **ignores** **filters** like ids, gender and other.


src/sql-queries/animal.ts
```diff
@@ -72,7 +72,8 @@ export const getAnimalsQuery = (
         ${table}.status,
         ${table}.image_url,
         ${table}.comments,
-        ${table}.mod_time`)
+        ${table}.mod_time,
+        (SELECT COUNT(*) FROM ${table}) AS total_count`)
```
AzisK, line 76: I am thinking that maybe the total count should be filtered by the same filters. WDYT?

AzisK (commented):
I am thinking that maybe the total count (for the filters) should be filtered by the same filters as the usual query. WDYT?

ignasposka:
Tests are failing can you take a look?

ignasposka (changes requested):
Looks good, but tests are failing and I'm getting `operator does not exist: gender = integer` while testing locally. Please fix that.

src/sql-queries/animal.ts
```diff
@@ -65,26 +65,42 @@ export const getAnimalsQuery = (
...
-        .leftJoin('breed AS b').on('ad.breed_id','b.id');
-    query = ids ? query.where($in(`${table}.id`, ids)) : query;
+        a.id,
+        a.name,
+        a.organization,
+        a.status,
+        a.image_url,
+        a.comments,
+        a.mod_time`)
+        .from(`${table} as a`);
```
AzisK, line 77: Since `AS` is SQL keyword, we could upper-case it here and in other places as well

src/sql-queries/animal.ts
```diff
@@ -65,26 +65,42 @@ export const getAnimalsQuery = (
...
-    query = reverse ? query.orderBy(`${table}.id DESC`) : query.orderBy(`${table}.id`);
-    query = limit != null ? query.limit(limit) : query;
+    const queryParams = query.toParams();
+
+    const selected = 'selected'
+    let pagination = select().from('selected as s');
+    pagination = cursor ? pagination.where(reverse ? lt('s.id', cursor) : gt('s.id', cursor)) : pagination;
+    pagination = reverse ? pagination.orderBy('s.id DESC') : pagination.orderBy('s.id');
+    pagination = limit != null ? pagination.limit(limit) : pagination;
+    pagination = pagination.crossJoin('total_count');
```
AzisK, line 95: I am not sure we need this cross join since we can simply select and avoid the join. I don't know if the cross join would affect performance. If yes, I suppose it should not affect significantly (tested locally having 4k+ rows). I am fine with a solution with cross joins as well.

AzisK (approved):

ignasposka (approved):

src/sql-queries/animal.ts
pergaliuke: I am leaving as it is for now.

Issue #244, opened 2021-05-24
Create table for Microchipping event details


AzisK:
**Goal**
Create a table for Microchipping event details with name `microchipping_event_details`

**Description** 
There already is a table of microchips. Therefore, this new, e.g. `microchipping_event_details` table could simply map these events to the microchip information table

Suggested `microchipping_event_details` table structure:
| Column name | Type |
| - | -|
| event_id | Int (FK to events table id column) |
| microchip_id | int (FK to animal_microchip table microchip_id column)  |

Please also add some fake data.
 
This issue depends on this issue https://github.com/pets-oss/pets-back/issues/250. Table `events` to store all events is first needed to then link (by foreign key) event details table to it.

Similar issue https://github.com/pets-oss/pets-back/issues/245

PR #264, opened 2021-06-06
#244 Create table for Microchipping event details

LukasKri:
Added new table `microchipping_event_details` and some fake data.

I had to make `microchip_id` column in `animal_microchip` table `UNIQUE`, because I get an error on Docker compose up command.

P. S. I think you made a typo in column name "microchipId" in the table structure.


AzisK (approved):
Initially event_id should have been a foreign key in `events` table. So this part is missing. However, we changed the architecture a little bit recently and we will rather have separate tables for the all the types of events. So it will need a little redo. I am marking this as approved and I will initialise the change soon. Thanks for the work.

PR #269, opened 2021-06-12
#268 Return isFavorite flag with Animal object

### LINKED ISSUE #268: Return isFavorite flag with Animal object
Check if an animal is marked as favorite by the current user and return a flag `isFavorite` with `Animal` object.

mantasimkus:
Supplemented `Animal` object with `isFavorite` flag which determines if an animal is marked as favorite by the current user. Resolves #268.

### DIFF
--- a/src/schema/resolvers/animal.ts
+++ b/src/schema/resolvers/animal.ts
@@ -24,7 +24,6 @@ import {
     updateAnimalRegistrationQuery,
 } from '../../sql-queries/animalRegistration';
 import { getStatusTranslationQuery } from '../../sql-queries/status';
-import { getFavoriteAnimalQuery } from '../../sql-queries/favoriteAnimal';
 
 const defaultLanguage: string = 'lt';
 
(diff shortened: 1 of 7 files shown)


AzisK (approved):
Maybe I would join the animal table with favorite table but other than that it look ok

mantasimkus:
@AzisK, as you noticed, currently `isFavorite` field is returned with a separate query (just like animal `registration`, `microchip`, and `details`). Are you suggesting to not follow the previous examples and join tables inside `getAnimal(s)Query`, or sth else?

AzisK:
I am thinking that this approach might be faster. However, as you pointed out, we have previous examples and it would be good to follow the same path. So I just gave food for thought but go ahead with this approach

src/schema/resolvers/animal.ts
```diff
@@ -325,6 +326,25 @@ const resolvers: IResolvers = {
...
+            }
+            
+            const dbResponse = await pgClient.query(
+                getFavoriteAnimalQuery({
+                    userId,
+                    animalId: id
+                })
+            );
+			
+            if (dbResponse.rows[0]) {
```
JustasBelevicius, line 343: could be: `return !!dbResponse.rows[0]`

JustasBelevicius (approved):
I agree with @AzisK, it probably would be nicer to just extend the animal query and model by adding the favourite flag (add logic to SQL - if user undefined/null, favourite always false). I believe an additional request for each animal is a bit unnecessary.

mantasimkus:
OK, as both of you suggested, I extended `getAnimal(s)` queries with `isFavorite` status. However, a few tests do not pass anymore due to undefined `user_id` used in the queries mentioned before. Would you suggest to hardcode a dummy `user_id` in tests?

JustasBelevicius:
Well done! Looks good from my point of view. @AzisK is this okay to merge for you?

AzisK:
Looks great

Issue #234, opened 2021-05-15
Return appropriate HTTP status code according to exception type.


ignasposka:
Setup solution which would capture exception and return appropriate HTTP status code for the user. Example of how to do it can be found in this blogpost: https://medium.com/@estrada9166/return-custom-errors-with-status-code-on-graphql-45fca360852
You are free to use your own solution, this is just an example.

So far we need to capture `ValidationError` and return HTTP 400 (Bad request).

PR #277, opened 2021-08-24
#234 return appropriate http status code according to exception type

pergaliuke:


pergaliuke:
After Migrating to Apollo server 3, this solution will not work since in V3 almost all plugin methods are async.

AzisK (approved):
Looking good!

Issue #276, opened 2021-08-18
Make createFavoriteAnimal not fail if the user already has that animal as favorite


AzisK:
**Goal**
Make createFavoriteAnimal not fail if the user already has that animal as favorite

**Description**
Simply return the information if the user already has the animal as favorite. Right now it returnas an exception of `duplicate key value violates unique constraint \"animal_favorite_pkey\"`

Code to refer to can be found here https://github.com/pets-oss/pets-back/blob/79a73532aed262cc87c05d0dedb7ab5da1b3d202/src/schema/resolvers/favoriteAnimal.ts#L43-L59

PR #283, opened 2021-08-31
#276 Make createFavoriteAnimal not fail if the user already has that animal as favorite

pergaliuke:


src/schema/resolvers/favoriteAnimal.ts
```diff
@@ -47,6 +48,18 @@ const resolvers: IResolvers = {
...
+            const dbResponseGetAnimal = await pgClient.query(
+                getFavoriteAnimalByIdQuery({
+                    userId,
+                    animalId
+                })
+            );
+            if (dbResponseGetAnimal.rows[0]) {
+                return dbResponseGetAnimal.rows[0];
+            }
+
```
AzisK, line 60: Nitpick: seems a lot of newlines

AzisK (approved):
Good job!

PR #301, opened 2021-10-09
Add possibility to filter colors by species, closes #298

### LINKED ISSUE #298: Add possibility to filter colors by species
**Goal**
Add possibility to filter colors by species

**Description**
It would be good to have an optional filter to filter colors by species. the filter, e.g. "speciesId" should be optional

**Expected behaviour**
Such GraphQL query
```graphql
{
  colors(language: "lt", speciesId: 11)
  {
    id
    value
  }
}
```

Should return all color entries for the species, e.g. (only sample):
```json
{
  "data": {
    "colors": [
      {
        "id": 83,
        "value": "Balta"
      },
      {
        "id": 84,
        "value": "Gelsva"
      }
    ]
  }
}
```

pergaliuke:


test/color.graphql.test.ts
```diff
@@ -28,4 +28,26 @@ describe('GraphQL color_translation integration tests', () => {
                 return done();
             });
     });
+
+    it('Returns colors translation in "lt" with all fields', (done) => {
```
AzisK, line 32: Maybe it could mention that species filter is applied?

